PR #1, opened 2016-10-15
Git Cli Module => Environments Module (Cli)

M3kH:
The proposal is to gentle give an introduction on environments rather than focusing on git-cli it self, since sometimes people might don't know whe context of execution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ If everyone can help out with elaborating on the different subjects that would b
 # What
 
 ## Git introduction (one week)
-  * publish your websites on Git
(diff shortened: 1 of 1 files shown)


Razpudding:
Hmm yeah I see where you're coming from but we were asked to do the exact opposite (teach CLI basics) by Maartje to try out if it works. If we find out today it's not a good idea we'll accept the merge.

Razpudding:
Ok it went pretty well in the beginning but as soon as people started making mistakes in the command line, there were too many different (And weird) problems to solve. So I think starting out with Github GUI is better.

PR #30, opened 2018-01-10
Updates for JavaScript 2 Class 13

remarcmij:
- Added code formatting fundamental
- Added explanation when to use functions
- Corrected broken links to old reviews
- Changes to reflect new curriculum ordering
- Changed some var's to let and const


mkruijt (approved):
Thank you @remarcmij for getting out all the vars. I really like the piece about "why we use functions". If you like you can add your robot assignments to week 3 and 4. I think it is a nice addition that they also work on a project next to the small coding assignments they have to do.

PR #37, opened 2018-02-12
Adding few more resources for promises

jalalalwani:


mkruijt:
hi @jalalalwani I think you have an old fork of the javascript repo, we made some changes in the schedule. Why don't you pull before you make a PR?

PR #39, opened 2018-02-27
Update promises.md

gijscor:
Added extra resources that students previously used to grasp the concept of Promises


mkruijt (approved):
cool, thanks @gijscor

PR #40, opened 2018-02-28
Updated/added fundamentals

remarcmij:
I have added a fundamental on the Event Loop and extended the promises fundamental. A minor correction to the map & filter fundamental, in that `reduce` does not return an array by default.

In JS3, Week 9 I have interpreted the topic of 'Code Flow' as an explanation of the Event Loop, hence the added fundamental. Please comment if you are of a different opinion.

As promises and the Event Loop are complicated subjects to explain I would really appreciate a critical review of these fundamentals.


fundamentals/map_filter.md
```diff
@@ -199,7 +199,9 @@ Result:
 
 ### Method chaining
 
-The methods **map()**, **filter()** and **reduce()** each return a new array. This makes it possible to chain these methods and create a 'pipeline' of operations, to be applied in sequence. Let's take the last example, but now filtering out only those array elements for which the name starts with a 'J':
+The methods **map()** and **filter()** each return a new array. This makes it possible to chain these methods and create a 'pipeline' of operations, to be applied in sequence. The **reduce** method can return anything, including an array. If a **reduce** method returns something other than an array it can only be included at the end of an array method chain.
```
jasongwartz, line 202: Nice clarification

fundamentals/promises.md
```diff
@@ -70,44 +81,61 @@ If you are only interested in the error case, you can pass `null` for the first
 somePromise.then(null, onRejected)
 ```
 
-or you can use a second method exposed by a promise object, which is just a short-hand for calling `then()` with `null` as its first argument:
+or you can use a second method exposed by a promise, which is just a short-hand for calling `then()` with `null` as its first argument:
```
jasongwartz: Good point here. I also find this line a bit confusing, especially if you didn't know the details of how `then()` works, might be worth further clarifying how `catch()` is used in combination with `then()`.

jasongwartz (approved):
Great refresh of the materials! I left of a couple comments, but nothing blocking - overall LGTM 👍

fundamentals/promises.md
remarcmij: I've added a `.then()` before the `.catch()` in the code snippet. Further down there is an explanation why `.catch()` should be at the end of the chain, but I realise at this point in the text it remains a bit vague.

jasongwartz:
Made a tiny change to clarify that sentence: fa0387b

GeorgeSapkin:
Maybe clarify that `catch` doesn't have to be the last method in the chain and can be used to handle errors in the middle. Example:

```js
Promise.reject(new Error('bad')).catch(() => 'good').then(console.log);

// outputs 'good'
```

PR #45, opened 2018-03-21
Fix Link in JS1 for Fundamentals

ChrisOwen101:
@mkruijt


mkruijt (commented):
thanks, missed that one

mkruijt (changes requested):
:)

mkruijt (approved):

PR #48, opened 2018-04-10
added note about lynda

wouterkleijn:


mkruijt (approved):
great work @wouterkleijn 😄

PR #53, opened 2018-09-04
Remove two sources of reading material

gijscor:
Both resources are very far from beginner friendly and clearly suppose previous coding experiences. 
Example of a sentence used in these resources:

"javaScript contains a standard library of objects, such as Array, Date, and Math, and a core set of language elements such as operators, control structures, and statements. Core JavaScript can be extended for a variety of purposes by supplementing it with additional objects"

As the level of English of some of our students is limited, and many have no background in programming, I think these resources are mostly just having students read information that they are not yet ready to process, and mostly will leave them confused. There are much better resources (that we have written ourselves) that explain it much better, and that we don't link to in the reading material of this week. (I have created a seperate pull request for this.

### DIFF
--- a/Week2/README.md
+++ b/Week2/README.md
@@ -25,9 +25,9 @@ Only watch the below chapters:
 2. Core Programming Syntax 
 3. Variables and Data Types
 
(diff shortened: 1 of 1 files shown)


remarcmij (approved):
Agreed.

M3kH:
I think that this was a mistake.

PR #56, opened 2018-09-18
corrections of outdated information

gijscor:
Step 1  made no sense, as they refer to previous javaScript exercises, whereas in week 1 they haven't done any JavaScript exercises yet. 
Step 2 also doesn't really makes sense as the students don't yet hand in their homework using pull requests, so commenting can only be done by making an issue, which is not the common way of commenting on code. I think it's much better to do this at a later stage when they make pull requests

### DIFF
--- a/Week1/MAKEME.md
+++ b/Week1/MAKEME.md
@@ -11,14 +11,13 @@ Topics discussed in class this week:
 
 _Deadline Monday_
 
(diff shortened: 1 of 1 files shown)


remarcmij (approved):
LGTM

Note that while it is possible to give feedback via an issue in each student's repo (they have one for their HTML/CSS homework; it doesn't require a PR), it is probably premature to expect that at this point.

PR #58, opened 2018-09-19
replace another link that is too hard to read

gijscor:
replaced 2 links ( from eloquent javaScript) that is too hard to read for beginners / non native english speakers. 
the first I replaced it with our internal Scope documentation, that currently wasn't part of the reading (and is talked about in Eloquent JS) 
the second I replaced with a clear video on expressions vs statements. I think in general it's good to be very specific about what each reading is exactly about, and what you should get understand out of it.

### DIFF
--- a/Week3/README.md
+++ b/Week3/README.md
@@ -32,10 +32,10 @@ Please watch the following parts of the course, [Programming Foundations Fundame
 - [Array](http://javascript.info/array)
 - [Loops](http://javascript.info/while-for)
 - [Functions](http://javascript.info/function-basics)
-- [Another Functions article](http://eloquentjavascript.net/03_functions.html)
+- [Scope](https://github.com/HackYourFuture/fundamentals/blob/master/fundamentals/scope.md)
 - [Objects](http://javascript.info/object)
 - [Conditions](http://javascript.info/ifelse)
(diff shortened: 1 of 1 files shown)


remarcmij (approved):
Thanks @gijscor, these are useful improvements. But note that we have more links to _Eloquent JavaScript_ in later modules, replacing the outdated _A Smarter Way..._. I think that at some point (not necessarily at this early stage) our students should be able to read more technically worded material such as _Eloquent JavaScript_, if they are to become self-supporting.

gijscor:
@remarcmij I agree that students will at some point will have to understand more complex ways of writing, but the question is when that is. I think its more important to make sure they understand the fundamentals well in simple language, before we start exposing them to more complex texts that refer to those fundamentals. My impression now is that when you expose students to these complex texts too early, the information is simply not digested in any form.

M3kH:
I’m afraid that these desicion shouldn’t be rush to be merge.

Replacing one reference to another one is a matter of common agreement and is especially valuable to take in count what other teachers of JavaScript module thinks about.

Merging this PR with haim to remove the references on Eloquent JavaScript for me is a not go; especially if we didn’t evaluate the topic it self.

gijscor:
@M3kH I understand your point, I will give it a bit more time for feedback next time.

PR #69, opened 2019-03-04
homework js week2

mohammadnaefaldarwesh:


Week2/MohammadDarwesh/mohammaddarwesh.js
```diff
@@ -0,0 +1,180 @@
...
+   };
+console.log(helloWorld3);
+console.log(helloWorld3.indonesian);
+console.log(helloWorld3.italian);
+console.log(helloWorld3.spanish);
+// end of task 1.
+
+
+// task 2:
+console.log("I'm awesome");
```
yash-kapila, line 36: Just for your information, you can also use escape characters(using backslash) like this:

Week2/MohammadDarwesh/mohammaddarwesh.js
```diff
@@ -0,0 +1,180 @@
+// for the first task i will do it with many ways just for practising.
+console.log ("Halo, dunia! ,Ciao, mondo! ,Hola, mundo!");
+
+let indonesian = "Halo, dunia!";
```
yash-kapila, line 4: I really liked how you tried to use different ways to print Hello world here. A point to note, however, is that you could have used `const` keyword instead of `let` while creating these variables. This is because you aren't re-assigning `indonesian`, `italian`, `spanish`, `helloWorld` etc variables anywhere.

If you remember from our class, always start using `const` for variables and switch to `let` if your program requires re-assigning a variable. 🙂

Week2/MohammadDarwesh/mohammaddarwesh.js
```diff
@@ -0,0 +1,180 @@
...
+console.log(helloWorld3.italian);
+console.log(helloWorld3.spanish);
+// end of task 1.
+
+
+// task 2:
+console.log("I'm awesome");
+
+// task 3:
+let x;
```
yash-kapila, line 39: This is correct. Here using a `const` keyword wouldn't have worked because we need to re-assign `x` to 32 🙂

Week2/MohammadDarwesh/mohammaddarwesh.js
```diff
@@ -0,0 +1,180 @@
...
+console.log("the value of my variable x will be: 32");
+console.log(x);
+x = 32;
+console.log("x will be :32 ");
+console.log(x);
+
+// task 4:
+let y = "Hello, Yash"
+console.log("the value of my string will be: Hello, Yash ");
+console.log(y);
```
yash-kapila, line 49: 👍

Week2/MohammadDarwesh/mohammaddarwesh.js
```diff
@@ -0,0 +1,180 @@
...
+console.log("the value of my string will be: Hiiii, Yash ");
+console.log(y);
+
+// task 5:
+let z = 7.25;
+console.log(z);
+let a = Math.round(z);
+console.log(a);
+console.log(a > z , z > a );
+if (z > a) {
```
yash-kapila, line 60: Aah. There is a bug here which you didn't notice because the code never reached the `else` block. You will notice it if you maybe did something like: `if (z < a)......`

Anyways, the bug is because of how scoping of variables work when created using `let` or `const` or `var` keyword. We will be covering that in our tomorrow's class. 🙂 

Other than that, you could have used `const` keyword for creating variables instead of `let` for `z` and `a`.

Week2/MohammadDarwesh/mohammaddarwesh.js
```diff
@@ -0,0 +1,180 @@
...
+    console.log(m);
+}
+else{ 
+    m = a;
+    console.log(m);}
+
+
+// task 6:
+let animals;
+console.log("the value of my array will be: cat, dog, hamster");
```
yash-kapila, line 71: Hi Muhammad, your `animals` array was supposed to be created as an empty array. However, like this, `animals` isn't assigned to any value and hence would be `undefined` when you log it.

You were needed to create an empty array i.e `const animals = [];`

Week2/MohammadDarwesh/mohammaddarwesh.js
```diff
@@ -0,0 +1,180 @@
...
+console.log(animals);
+let myFavoriteAnimals = ["cat", "dog", "hamster"];
+console.log(myFavoriteAnimals);
+myFavoriteAnimals[3] = "baby pig"
```
yash-kapila, line 75: Even though this works just fine, if you are required to add an element to an array possibly at the end of current array, you should try to use the `push` method we discussed during our class.

That's because although we know the current size of the array, we can't be really sure that it would always contain 3 elements. What if it were supposed to contain 5 elements or we didn't know what the length of the array is. In such case, doing `myFavouriteAnimals[3] = "baby pig"` wouldn't work.

`myFavouriteAnimals.push("baby pig");` would be safer in such a case, right?

Week2/MohammadDarwesh/mohammaddarwesh.js
```diff
@@ -0,0 +1,180 @@
...
+console.log(myFavoriteAnimals);
+
+// task 7:
+let myString = "this is a test";
+console.log(myString);
+let theLengthOfMyString = myString.length;
+console.log(theLengthOfMyString);
+/* i did the previous two steps
```
yash-kapila, line 83: This is correct. I understand why you did it like that. However, please try to use `const` always as a first preference for creating variables. 🙂

Week2/MohammadDarwesh/mohammaddarwesh.js
```diff
@@ -0,0 +1,180 @@
...
+let favoriteAnimals = ["cat", "dog"];
+let fact = 1 < 2;
+console.log(name);
+console.log(age);
+console.log(favoriteAnimals);
+console.log(fact);
+console.log("the type of my variables is: string, number, array, boolean");
+console.log(typeof name);
+console.log(typeof age);
+console.log(typeof favoriteAnimals);// i do not know why it showing as an object
```
yash-kapila, line 100: We will cover this tomorrow 🙂 Or did you try to find it out on google? 😉

Week2/MohammadDarwesh/mohammaddarwesh.js
```diff
@@ -0,0 +1,180 @@
...
+console.log(myString);
+let theLengthOfMyString = myString.length;
+console.log(theLengthOfMyString);
+/* i did the previous two steps 
+just because they are in the task as two steps 
+but you can  do it just with the following step */
+console.log(myString.length);
+
+// task 8:
+let name = "mike";
```
yash-kapila, line 89: `const` please

Week2/MohammadDarwesh/mohammaddarwesh.js
```diff
@@ -0,0 +1,180 @@
...
+console.log(6/0 > 10/0);
+if (6/0 === 10/0){
+    console.log("yes i can");
+}
+else {
+    console.log("no i can not");
+};
+
+
+
```
yash-kapila, line 134: Could you please ensure that there aren't any such un-needed spaces next time onwards in your PR 🙂

Week2/MohammadDarwesh/mohammaddarwesh.js
```diff
@@ -0,0 +1,180 @@
...
+x = 7;
+x = x % 3;
+console.log(x);
+// x will be (1).
+console.log("5 % 1 = 0  because 1 divides 5 (or any other number perfectly)");
+console.log("10 % 3 = 1  Attempting to divide 10 by 3 would leave remainder as 1");
+x = 9;
+x = x % 5;
+console.log("i think x will be 4 " );
+console.log(x);
```
yash-kapila, line 117: Correct

yash-kapila (approved):
Hi @mohammadnaefaldarwesh - I have reviewed your homework and am approving the same. However, I see that you have used `let` keywords everywhere to create variables. If you remember from our last week's class, you should always strive for using `const` keyword while creating variables and switch to `let` only if you end up in a situation in your program, where you need to re-assign your variable. Could you please ensure that you use `const` the next time onwards?

Week2/MohammadDarwesh/mohammaddarwesh.js
yash-kapila: Also, you could also use the `Math.max` method here instead of writing the `if-else` block. Please find some more information about this method here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max

wouterkleijn:
Dear @mohammadnaefaldarwesh Well-done! Thanks for sending in the homework in time and doing a great job.

It would be nice if you can respond to @yash-kapila 's comments !

Keep up the good work.

Week2/MohammadDarwesh/mohammaddarwesh.js
mohammadnaefaldarwesh: Thank you for your feed back .
I know already about (backslash).... but in general most of the questions are not specific....i think it will be better if the question is (can you find many ways to solve the error? ).

Week2/MohammadDarwesh/mohammaddarwesh.js
mohammadnaefaldarwesh: yes it is my mistake .....i get it wrong in the class but now it is clear ....thank you.

Week2/MohammadDarwesh/mohammaddarwesh.js
mohammadnaefaldarwesh: thank you for explain it to me in the class .....and i am start using strict mode to avoid this kinds of bug.
about the Math.max ......if the purpose from the question to use  Math.max i think also will be better if the question more specific .

Week2/MohammadDarwesh/mohammaddarwesh.js
mohammadnaefaldarwesh: same as Math.max .....i think also will be better if the if the question more specific .

Week2/MohammadDarwesh/mohammaddarwesh.js
mohammadnaefaldarwesh: yes sure ....my mistake !!!

PR #71, opened 2019-03-04
homework JawaScript1 Week2

RamazanInce:


Week2/ramazanInce.js
```diff
@@ -0,0 +1,158 @@
+console.log(`
+Question 1`);
```
yash-kapila, line 2: Hi Ramazan, was there a reason you preferred to use back-ticks here and split the string to two lines?

Week2/ramazanInce.js
```diff
@@ -0,0 +1,158 @@
...
+
+console.log(`
+Question 3`);
+// 3. Declare a variable x and initialize it with
+// an integer, using these exact steps:
+// 3.1 First, declare your variable x (do not initialize it yet).
+// 3.2 Add a console.log statement that explains in
+// words what you think the value of x is, like in this example:
+
+var x;
```
yash-kapila, line 27: Hi Ramazan, you should try to avoid using `var` as much as possible. With the new JavaScript standard(ES6), we should try to use `let/const`. Very rarely if not never would you need the use of `var` keyword 🙂

Week2/ramazanInce.js
```diff
@@ -0,0 +1,158 @@
...
+Priviyet Mir
+Marhaban Dunya
+Hallo Wereld
+`);
+
+console.log(`
+Question 2`);
+// 2. Consider the following code:
+
+console.log("I'm awesome");
```
yash-kapila, line 17: Just for your information, you can also use escape characters(using backslash) like this:

Week2/ramazanInce.js
```diff
@@ -0,0 +1,158 @@
...
+console.log('the value of my variable x will be : whateverYouThinkItWillLog');
+console.log(x);
+x = 41;
+console.log('the value of x is my age');
+console.log(x);
+
+console.log(`
+Question 4`);
+// 4. Declare a variable y and assign a string to it.
+var y = `Leersum`;
```
yash-kapila, line 37: You can use `let` here instead of `var` 🙂

Week2/ramazanInce.js
```diff
@@ -0,0 +1,158 @@
...
+console.log(`
+Question 5`);
+//5. How do you round the number 7.25, to
+//the nearest integer (i.e., whole number)?
+
+let z = 7.25;
+console.log(z);
+let a = Math.round(z);
+console.log(a);
+let b = Math.max(a, z);
```
yash-kapila, line 53: I really liked your usage of `Math.max` to find out the max of `a` and `z` 🙂 

But you could have used `const` keyword here instead of `let` as we are never reassigning `a`, `b` or `z`, isn't it?

Week2/ramazanInce.js
```diff
@@ -0,0 +1,158 @@
...
+console.log(`
+Question 6`);
+// 6. Arrays
+
+let cities = ['', '', ''];
+console.log(`the array is the list of some of capital cities in the world`);
+console.log(cities);
+let myFavoriteAnimals = ['dog', 'cat', 'horse', 'rapid', 'bird', 'camel'];
+console.log(myFavoriteAnimals);
+myFavoriteAnimals.push(`baby pig`);
```
yash-kapila, line 65: thumbs-up for using the `push` method

Week2/ramazanInce.js
```diff
@@ -0,0 +1,158 @@
...
+let a = Math.round(z);
+console.log(a);
+let b = Math.max(a, z);
+console.log(b);
+
+console.log(`
+Question 6`);
+// 6. Arrays
+
+let cities = ['', '', ''];
```
yash-kapila, line 60: `const` here as well please 😉 

Also, when you hear the term `empty-array` in JavaScript, it actually means creating an array with 0 elements i.e `const cities = [];`.

Week2/ramazanInce.js
```diff
@@ -0,0 +1,158 @@
...
+console.log(genderMale);
+let typeOfName = `string`;
+let typeOfAge = `number`;
+let typeOfWorkPermit = 'undefined';
+let typeOfGenderMale = `boolean`;
+console.log(`I think type of name is ${typeOfName}`);
+console.log(`I think type of age is ${typeOfAge}`);
+console.log(`I think type of workPermit is ${typeOfWorkPermit}`);
+console.log(`I think type of genderMale is ${typeOfGenderMale}`);
+if (typeof name === typeOfName) {
```
yash-kapila, line 97: Hi Ramazan, I think you mis-understood requirement **8.5**. You are supposed to equate the types of different variables which you have created i.e name, age, workPermit and genderMale instead of equating them with their `typeXxx` variables.

Something like:

```
if (typeof age === typeof name) {
  console.log('SAME TYPE');
} else {
  console.log('DIFFERENT TYPE');
}
```

And so on

Week2/ramazanInce.js
```diff
@@ -0,0 +1,158 @@
...
+t = t % 15;
+console.log(t);
+
+console.log(`
+Question 10`);
+console.log(`Yes`);
+let myPersonelData = ['Ramazan', 'Ince', 41, 'Leersum', 3956, , null, true];
+console.log(myPersonelData);
+console.log(
+  typeof myPersonelData[0],
```
yash-kapila, line 148: This is an interesting thing. This actually shows why we have `loops` in every language which we are going to cover tomorrow. 🙂

yash-kapila (approved):
Hi @rincemath - I have reviewed your homework and am approving the same. However, please take care of using `let` and `const` over `var` next time onwards. I won't be lenient then. 😉 

Also, you got requirement **8.5** a tad wrong. Please take care of reading the instructions carefully during the homework. Understanding the problem statement or requirements is an equally important step as writing code to solve it. I hope you are careful the next time onwards. See you tomorrow!

Week2/ramazanInce.js
RamazanInce: Yes, when you run the code I want to put an empty line between the questions.

Week2/ramazanInce.js
RamazanInce: Ok

Week2/ramazanInce.js
RamazanInce: Yes, I know that. I believe most of our class mate may use escape character then to have a different answer I did in that way.

Week2/ramazanInce.js
RamazanInce: I did not remember but for some conseles let did not work. That is why I used var.

Week2/ramazanInce.js
RamazanInce: Through out the homework I try to use different letter not to have a problem since it was in one page. But for the followings works I will do in that way. Thanks

Week2/ramazanInce.js
RamazanInce: That was the point in the question which was not clear enough for me. Since 'plurality' may have a special meaning.

Week2/ramazanInce.js
RamazanInce: Yes I complety miss understood

because of the following statements, I thought you are asking us to compare for the same variable type of what we think and what it is.

8.3 Now write a console.log statement wherein you first explain in words what you think the type of your variables is.

RamazanInce:
Hi Yash, I would like to tell my thanks and appreciate you for giving feedback for each question separately.
It assist me a lot and I am happy to have teacher like you.

wouterkleijn:
Thanks for being so responsive and follow-ing up on the feedback. Well done @rincemath !

RamazanInce:
> Thanks for being so responsive and follow-ing up on the feedback. Well done @rincemath !

Actually you are the one deserve the thanks you also follow us whether are we doing things which we should do or not

PR #72, opened 2019-03-05
Muhannad homework

Muhannad1989:
this is my home work


Week2/muhannad.js
```diff
@@ -0,0 +1,144 @@
+// say hello world in 3 language
+console.log('Hello World'); //English
+console.log('Hola Mundo'); //Spanish
+console.log('Bonjour le monde'); //French
+
+//SyntaxError: Find a solution for this error.
+console.log("I'm awesome");
```
yash-kapila, line 7: Just for your information, you can also use escape characters(using backslash) like this:

Week2/muhannad.js
```diff
@@ -0,0 +1,144 @@
...
+//3.2 Add a console.log statement that explains in words what you think the value of x is, like in this example:
+console.log(`the value of my variable x will be undefined because it dose not have any value yet`);
+//3.3 Add a console.log statement that logs the value of x.
+console.log(x);
+//3.4 Now initialize your variable x with an integer.
+x = 11;
+//3.5 Next, add a console.log statement that explains what you think the value of x is.
+console.log('the value of x will be 11');
+//3.6 Add a console.log statement that logs the value of x.
+console.log(x);
```
yash-kapila, line 20: 👍

Week2/muhannad.js
```diff
@@ -0,0 +1,144 @@
...
+console.log(z);
+//5.3 Declare another variable a that has the value of z but rounded to the nearest integer.
+let a = z;
+//5.4 console.log a.
+console.log(Math.round(a));
+//5.5 So now we have z and a find a way to compare the two values and store the highest of the two in a new variable.
+//i think there are many way's to compare between 2 'built-in functions' or more value's. this is in a different way one of them.
+//5.6 console.log the highest value.
+if (a > z) {
+  console.log('bigger than b', a);
```
yash-kapila: Hi Muhannad, even though you check for the bigger value between `a` and `z`, you missed this requirement: **store the highest of the two in a new variable.**

How would you do it? Also, just wondering if you checked the in-built `Math.max` function from JavaScript. If not, you can check the documentation of it on `mdn`.

Week2/muhannad.js
```diff
@@ -0,0 +1,144 @@
...
+console.log(y);
+//4.3 Now assign a new string to the variable y.
+y = 'i am studying javascript';
+//4.4 Just like what you did before write a console.log statement that explains in words what you think will be logged to the console.
+console.log('the value of x will be: i am studying javascript');
+//Now console.log y again.
+console.log(y);
+//How do you round the number 7.25, to the nearest integer (i.e., whole number)?
+//5.1 Declare a variable z and assign the number 7.25 to it.
+let z = 7.25;
```
yash-kapila: You can use `const` here instead of `let`, can't you? Since you aren't re-assigning `z` or `a` anywhere else in your code 😉

Week2/muhannad.js
```diff
@@ -0,0 +1,144 @@
...
+//i think there are many way's to compare between 2 'built-in functions' or more value's. this is in a different way one of them.
+//5.6 console.log the highest value.
+if (a > z) {
+  console.log('bigger than b', a);
+} else {
+  console.log('z is bigger than a', z);
+}
+///////////Arrays/////////////
+//6.1 Declare an empty array. Make sure that the name you choose indicates 'plurality', because an array is capable of containing more than one element. (See Naming conventions).
+let arr1 = [, , ,];
```
yash-kapila: Hi Muhannad - you can also create an empty array like this: `const arr = [];`. It is similar to what you have already.

Also, please focus on using `const` keywords for creating variables wherever you can. 🙂

Week2/muhannad.js
```diff
@@ -0,0 +1,144 @@
...
+//6.2 Write a console.log statement that explains in words what you think the value of the array is.
+console.log('the value of arr1 will be undefined or empty');
+//6.3 console.log your array.
+console.log(arr1);
+//6.4 Create an array that has your favorite animals inside (see if you can find a good name that exactly describes what this variable will hold).
+let favoriteAnimals = ['rabbit', 'cow', 'dog', 'cat'];
+//6.5 Log your array.
+console.log(favoriteAnimals);
+//6.6 Add a statement that adds Daan's favorite animal ('baby pig') to the existing array.
+favoriteAnimals.push('baby pig');
```
yash-kapila, line 66: 👍 for the use of `push` function

Week2/muhannad.js
```diff
@@ -0,0 +1,144 @@
...
+//show latest Array:information
+for (let i = 0; i < information.length; i++) {
+  console.log(information[i]);
+}
+//10.2 Can you compare infinities? (Not in Eyad's world) - does 6/0 === 10/0? How can you test this?
+//10.3 Add console.log statements to the above program in which you show that you understand the concepts (just like you've done in the above assignments).
+let number1 = 6 / 0,
+  number2 = 10 / 0;
+console.log(number1, number2);
+console.log(number1.typeOf, number2.typeOf); //can not divide never by 0 because it give infinity
```
yash-kapila: Hi @Muhannad1989 , what are you trying to do here in line 127. Can you please elaborate?

yash-kapila (commented):
Hi @Muhannad1989  - I have reviewed your PR but have one question around the last question in the homework. Can you please have a look?

Also, I would like to re-iterate that you should try to use `const` keyword wherever you can and should only use `let` if you have the need to re-assign your variable. `var` should if not never then rarely used.

Week2/muhannad.js
Muhannad1989: Hi yash,
i was trying to make some challenge without built-in function, but i have changed it to the following code:
const maxValue = Math.max(a, z);
console.log(maxValue);

yash-kapila: Thank you

yash-kapila (approved):

wouterkleijn:
Well done @Muhannad1989 Thanks for pushing your improvements based on the feedback of Yash.

For the next time it would be nice if you can respond to @yash-kapila 's comments here as well.

Keep up the good work.

PR #73, opened 2019-03-05
Akif homework

Akif2709:


Week2/akif.js
```diff
@@ -0,0 +1,108 @@
+// Part 1 Hello World
+console.log('Hello World'); // English
+console.log('Merhaba Dunya'); // Turkish
+console.log('Hallo Wereld'); // Dutch
+console.log('مرحبا بالعالم'); // Arabic
+
+// Part 2 Error  -----------------------------------------
+console.log("I'm awesome");
```
yash-kapila, line 8: Just for your information, you can also use escape characters(using backslash) like this:

Week2/akif.js
```diff
@@ -0,0 +1,108 @@
...
+
+// Part 3 Console.log(x)   --------------------------------
+let x;
+console.log('the value of my variable x will be: numberOfTheStudents');
```
yash-kapila: You were supposed to replace `numberOfTheStudents` with the value which you expect it should have been i.e `undefined` or `16` or something like that. You should be doing it before you run the code in browser or VS Code terminal. This is needed to ensure that you understand how JS works. 🙂 

Can you please correct it?

Week2/akif.js
```diff
@@ -0,0 +1,108 @@
...
+x = 16;
+console.log('the value of my variable x will be: numberOfTheStudents');
+console.log(x);
+
+// Part 4  String y  ---------------------------------------
+let y = 'I love you JavaScript';
+console.log('the value of my variable y will be: feelingAboutJavaScript');
+console.log(y);
+y = 'I will learn you';
+console.log('the value of my variable y will be: myPlanAboutJavaScript');
```
yash-kapila: 😆

Week2/akif.js
```diff
@@ -0,0 +1,108 @@
...
+console.log(y);
+
+// Part 5  Round a number  ---------------------------------
+let z = 7.25;
```
yash-kapila: Can use `const` here instead of `let`, isn't it? Because we don't reassign `z` or `a` or `b`.

Week2/akif.js
```diff
@@ -0,0 +1,108 @@
...
+console.log(z);
+let a = Math.round(z);
+console.log(a);
+let b = Math.max(z, a);
```
yash-kapila: 👍

Week2/akif.js
```diff
@@ -0,0 +1,108 @@
...
+
+// Part 8  typeof  -------------------------------------------
+let k = 42;
+let l = 'Hello World';
+let m;
+let n = true;
+console.log(k, l, m, n);
+console.log('the types of my variable accordingly are: number, string, undefined, boolean');
+console.log(typeof k, typeof l, typeof m, typeof n);
+if (typeof k != typeof l) {
```
yash-kapila: Ah. If you remember from our class last week, we should always strive for using strict equality checks i.e `===` or `!==` instead of `==` or `!=` respectively. Can you please update this?

Week2/akif.js
```diff
@@ -0,0 +1,108 @@
...
+  console.log(y + ' is a odd number');
+}
+
+// Part 10 --------------------------------------------------------------
+let arr = [13, 'I am a programmer', undefined, null];
+console.log(arr);
+
+function arrayTypeChecker(b) {
+  for (let i = 0; i < b.length; i++) {
+    if (typeof b[i] !== typeof b[i + 1]) {
```
yash-kapila, line 98: Hi Akif - although your logic of checking for types of different array elements is correct this way, it has got one small bug.

You are running the loop from 0 until `i` is less than array length which in our case is 4. That means, our last iteration would have `i = 3/b[i] = null` and then when execution reaches line 98, it would try to equate `typeof b[3]` with `typeof b[4]` which doesn't exist and also goes against what we wanted to actually do.

This can be fixed easily if you ran the loop until `i < b.length - 1`. This way, our last iteration would compare `typeof b[2]` and `typeof b[3]` and would then exit out from the loop.

yash-kapila (changes requested):
Hi @Akif2709 - I have reviewed your PR but can't approve it right now as I would like you to look into a couple of comments which I have added. Can you please update your JS file and push the changes again to your branch? 

Please let me know if you have any questions around it.

Akif2709:
Hi @yash-kapila. I made the corrections according to your review. Thanks for your elaborated review 👍

Week2/akif.js
```diff
@@ -0,0 +1,108 @@
...
+console.log('Merhaba Dunya'); // Turkish
+console.log('Hallo Wereld'); // Dutch
+console.log('مرحبا بالعالم'); // Arabic
+
+// Part 2 Error  -----------------------------------------
+console.log("I'm awesome");
+
+// Part 3 Console.log(x)   --------------------------------
+let x;
+console.log('the value of my variable x will be: 16');
```
yash-kapila: Hi Akif - the value of `x` here would be `undefined` because we haven't assigned it a value on line 11. It's only a declaration. It is only at line 14 when we assign `x` with a value. 🙂

Week2/akif.js
```diff
@@ -0,0 +1,108 @@
...
+
+// Part 8  typeof  -------------------------------------------
+let k = 42;
+let l = 'Hello World';
+let m;
+let n = true;
+console.log(k, l, m, n);
+console.log('the types of my variable accordingly are: number, string, undefined, boolean');
+console.log(typeof k, typeof l, typeof m, typeof n);
+if (typeof k !== typeof l) {
```
yash-kapila, line 57: 👍

yash-kapila (approved):
Hi @Akif2709 - I have reviewed your changes but you had one small mistake still. Please have a look at my comment. I am approving the PR anyways. Good luck for this week's homework. 🙂

wouterkleijn:
Dear @Akif2709 Thanks for pushing your improvements and responding to Yash's feedback. Can you please push the last few improvements so we can approve of your PR? Thanks in advance.

PR #75, opened 2019-03-06
homeworkjavascript

alaashhadat:


Week2/alaa.js
```diff
@@ -0,0 +1,97 @@
+//1
+console.log('Hallo', 'wereld!'); // Dutch
+console.log('hola', 'mundo!'); //spanish
+console.log('مرحبا', 'العالم!'); //Arabic
+//2
+console.log("I'm awesome"); // should use two quotations instead of one
```
yash-kapila, line 6: Just for your information, you can also use escape characters using backslash i.e `\` like: `console.log('I\'m awesome');`. But the way you did is also correct.

Week2/alaa.js
```diff
@@ -0,0 +1,97 @@
...
+//3
+var x;
+console.log('the value of my variable x will be: undefined');
+x = 7; // initialize the value of x
+console.log('the value of my variable x will be 7');
+console.log(x);
```
yash-kapila, line 12: This is correct but if you remember from our class, we should always strive to use `const` keyword first followed by `let` and then `var`. You should try to avoid `var` keyword as much as possible. In this particular case, using `const` wouldn't work because we need to re-assign `x = 7` in line 10. Thus, you can use `let x;` instead of `var`. 🙂

Week2/alaa.js
```diff
@@ -0,0 +1,97 @@
...
+//4
+var y;
+y = 'hyf';
```
yash-kapila, line 15: Two things here please:
1. You can use `let` keyword here instead of `var`.
2. You can simply combine the two statements together i.e do `let y = 'hyf'`;

Week2/alaa.js
```diff
@@ -0,0 +1,97 @@
...
+var y = 'code';
+console.log('the value of y will be :code');
+console.log(y);
+//5
+var z = 7.25;
+console.log(z);
+var a = Math.floor(z);
+console.log(a);
+var max = Math.max(z, a);
+console.log(max);
```
yash-kapila, line 27: Again, please try to use `const` keyword always first followed by `let` and then very rarely should you use `var` keyword. Also, I would like you to find out the difference between `Math.floor` and `Math.round` 😉

Week2/alaa.js
```diff
@@ -0,0 +1,97 @@
...
+//6
+var myLibraries = [];
+console.log('the value of the the array is []');
+var myFavoriteAnimals = ['dogs', 'elephants', 'turtles', 'cats'];
+console.log(myFavoriteAnimals);
+var newLength = myFavoriteAnimals.push('baby pig');
```
yash-kapila, line 33: 👍

Week2/alaa.js
```diff
@@ -0,0 +1,97 @@
...
+console.log(newLength);
+//7
+let myString = 'this is a test';
+console.log(myString);
+myString.length;
```
yash-kapila, line 38: You don't need this line. If you remember from our class today, this is just an expression and not a statement. The computer doesn't know what to do with this line except evaluate the length of `myString`. If you want, you can convert this to a statement like `const len = myString.length;` and then do `console.log(len);` in the next line 🙂

Week2/alaa.js
```diff
@@ -0,0 +1,97 @@
...
+console.log(max);
+//6
+var myLibraries = [];
+console.log('the value of the the array is []');
+var myFavoriteAnimals = ['dogs', 'elephants', 'turtles', 'cats'];
+console.log(myFavoriteAnimals);
+var newLength = myFavoriteAnimals.push('baby pig');
+console.log(newLength);
+//7
+let myString = 'this is a test';
```
yash-kapila, line 36: This can be a `const` variable

Week2/alaa.js
```diff
@@ -0,0 +1,97 @@
...
+console.log('the value of my const a is :4');
+const b = 'dog';
+console.log('the value of my variable animal is :dog');
+const c = true;
+console.log('the value of my variable age28 is : true');
+const d = [1, 2, 3, 4];
+console.log('the value of variable a is: number');
+console.log('the value of variable b is : string ');
+console.log('the value of c variable is : boolean');
+console.log('the value of variable d is :array ');
```
yash-kapila, line 51: Are you sure this is an `array`? 🙂

Week2/alaa.js
```diff
@@ -0,0 +1,97 @@
...
+const typeA = 4;
+typeof typeA; //number
+const typeB = 'dog';
+typeof typeB; //string
+const typeC = true;
+typeof typeC; // boolean
+const typeD = [1, 2, 3, 4];
+typeof typeD; // number
```
yash-kapila, line 59: Again, simply typing `typeof typeD` here doesn't really mean anything because we don't do anything with this expression. It would make more sense if you did something like: `console.log(typeof typeD);` because now you would be printing the value of your expression to console.

Week2/alaa.js
```diff
@@ -0,0 +1,97 @@
...
+b = b % 2;
+console.log(1);
+//10
+var arr = [1, 'dog', 32, 'baby']; // you can add different types of values in one array.
+// you can compare   infinities
+var x = "Not in Eyad's world";
+var b = 6 / 0;
+var c = 10 / 0;
+if (b === c) {
+  console.log(true); // it shows true then you can compare infinities
```
yash-kapila, line 92: 👍

Week2/alaa.js
```diff
@@ -0,0 +1,97 @@
...
+console.log('the value of x is :1 ');
+var y = 8;
+y = y % 6;
+console.log(2);
+var z = 4;
+z = z % 3;
+console.log(1);
+var b = 11;
+b = b % 2;
+console.log(1);
```
yash-kapila, line 84: 👍

Week2/alaa.js
```diff
@@ -0,0 +1,97 @@
...
+console.log(typeC);
+console.log(typeD);
+if (typeA == typeB) typeA == typeC || typeA == typeD;
+{
+  console.log('same type');
+}
+if (typeA != typeB) typeA != typeC || typeA != typeD;
+{
+  console.log('Not the same type');
+}
```
yash-kapila, line 71: Hi @alaashhadat - there is a syntax problem here i.e you have misplaced the `)` bracket. Also, please try to use triple equals syntax i.e `===` instead of double equals `==` whenever you are doing a comparison. Can you please correct this? Let me know if you have any questions.

yash-kapila (changes requested):
Hi @alaashhadat - I have reviewed your PR but would like you to look at my comments for exercise **8** and address the suggestions. Also, please try to use `const` or `let` as much as possible and very rarely should you use `var` for declaring variables. I would be happy if you updated them as well. Otherwise, your PR is fine. Please drop me a message on Slack if you have any questions. 🙂

wouterkleijn:
Dear @alaashhadat Thanks for all your efforts. Would you mind addressing @yash-kapila's comments so we can approve your PR?

PR #76, opened 2019-03-06
orhan homework

OrhanGoc:


orhan.js
```diff
@@ -0,0 +1,102 @@
+//1. Part of Homework
+
+console.log('Merhaba, Dunya!\nHello, World!\nHallo, Wereld!');
```
yash-kapila, line 3: 👍

orhan.js
```diff
@@ -0,0 +1,102 @@
...
+
+//2. Part of Homework
+
+console.log('I\'m awesome');
+//this also works console.log("I'm awesome");
```
yash-kapila, line 8: 👍 😆

orhan.js
```diff
@@ -0,0 +1,102 @@
...
+
+//3. Part of Homework
+
+let x;
+console.log('the value of my variable x will be: undefined');
+console.log(x);
+x = 10;
+console.log('the value of my variable x will be: 10');
+console.log(x);
```
yash-kapila, line 17: Correct

orhan.js
```diff
@@ -0,0 +1,102 @@
...
+
+//4. Part of Homework
+
+let y = 'hello';
+console.log('the value of my variable y will be: hello');
+console.log(y);
+y = 'hi';
+console.log('the value of my variable y will be: hi');
+console.log(y);
```
yash-kapila, line 26: I like that you used `let` here instead of `var.` 🙂

orhan.js
```diff
@@ -0,0 +1,102 @@
...
+
+//5. Part of Homework
+
+let z = 7.25;
+console.log('the value of z is: ' + z);
+let a = Math.round(z);
+console.log('the value of a is: ' + a);
+let highestValue = a < z ? z : a;
+console.log('the highest value is: ' + highestValue);
```
yash-kapila, line 35: Two things here please:
1. You can use `const` keyword instead of `let` to create `z`, `a` and `highestValue` as we are not re-assigning any of those variables, right?
2. I would like you to learn about the difference between `Math.floor` and `Math.round` methods 🙂

orhan.js
```diff
@@ -0,0 +1,102 @@
...
+
+//6. Part of Homework
+
+let favoriteAnimals = [];
+console.log('the value of the array will be: []');
+console.log(favoriteAnimals);
+favoriteAnimals = ['bird', 'cat'];
+console.log(favoriteAnimals);
+favoriteAnimals[2] = 'baby pig';
```
yash-kapila, line 44: Even though this works just fine, it would be better if you used the `push` method i.e `favoriteAnimals.push("baby pig");`. This will be useful because you wouldn't know always for certain what the length of your array would be and then if you want to add an element to it, it would be hard to do it using the current way as you do. In such cases, `push` method would come in handy as it will add an element to the end of the array.

orhan.js
```diff
@@ -0,0 +1,102 @@
...
+console.log(favoriteAnimals);
+
+//7. Part of Homework
+
+let testString = 'this is a test';
+console.log(testString);
+let stringLength = testString.length;
+console.log(stringLength);
```
yash-kapila, line 52: Can use `const` here instead of `let` 🙂

orhan.js
```diff
@@ -0,0 +1,102 @@
...
+
+//8. Part of Homework
+
+let myNumber = 28;
+let myString = 'sea'
+let myBoolean = true;
+let myArray = [1, 2, 3, 4, 5];
```
yash-kapila, line 59: Can use `const` here again instead of `let` as we are not re-assigning any of these variables, right? 🙂

orhan.js
```diff
@@ -0,0 +1,102 @@
...
+
+let m = 19;
+m = m % 5;
+console.log('the value of m will be 4');
+console.log(m);
+
+let p = 37;
+p = p % 13;
+console.log('the value of p will be 11');
+console.log(p);
```
yash-kapila, line 91: 👍

yash-kapila (approved):
Hi @OrhanGoc - I have reviewed your PR and am approving the same. However, I do have a coupe suggestions. I see that you avoided using `var` keyword which is good but you can still use `const` keywords on a couple occasions where you use `let` right now. Also, there is one comment around using the inbuilt `push` method of an array instead of addressing an array with its index. Could you have a look please? 🙂

wouterkleijn:
Dear @OrhanGoc  Well-done! Thanks for sending in the homework in time and doing a great job.

It would be nice if you can respond to @yash-kapila 's comments and make a few more improvements!

Keep up the good work.

PR #77, opened 2019-03-06
Nadin- Homework - week1

Nadoshka:


Week1/homework/nadin.js
```diff
@@ -0,0 +1,111 @@
+'use strict';
+let greetingIndonesian = 'Ciao';
+let greetingArabic = 'مرحبا ';
+let greetingSpanish = 'Halo';
+ console.log(greetingIndonesian, greetingArabic, greetingSpanish);
```
yash-kapila, line 5: Hi Nadin, you can use the `const` keyword here instead of `let`. This is because we are not re-assigning any of the greeting variables, right? 🙂

Week1/homework/nadin.js
```diff
@@ -0,0 +1,111 @@
...
+
+console.log("I'm awesome");
+
+
+let x;
+console.log("The value of my variable number will be: undefined");
+console.log(x);
+x = 1;
+console.log("The value of my variable number will be: 1");
+console.log(x);
```
yash-kapila, line 16: 👍

Week1/homework/nadin.js
```diff
@@ -0,0 +1,111 @@
...
+
+let y = "the first string";
+console.log("the value of my variable y will be: the first string");
+console.log(y);
+y = "the second string";
+console.log("the value of my variable string1 will be: the second string");
+console.log(y);
```
yash-kapila, line 23: 👍

Week1/homework/nadin.js
```diff
@@ -0,0 +1,111 @@
...
+
+console.log(Math.round(7.25));
+// expected output: 8
```
yash-kapila: Are you sure this is going to be printed as 8? Can you please check? Also, I would like if you tried to find out the difference between `Math.floor` and `Math.round` 🙂

Week1/homework/nadin.js
```diff
@@ -0,0 +1,111 @@
...
+
+let z = 7.25;
+console.log(z);
+
+let a = 8;
+console.log(a);
+
+console.log(z !== a);
+console.log(z == a);
+console.log(z <= a);
```
yash-kapila: Hi Nadin, I think you perhaps mis-understood the requirement here. We want to do this:
`5.3 Declare another variable a that has the value of z but rounded to the nearest integer.`
`5.5 So now we have z and a find a way to compare the two values and store the highest of the two in a new variable.`.

That means, instead of hardcoding `let a = 8;`, you need to do this `const a = Math.round(z);`.

After this, you need to compare the two variables `a` and `z` and store the maximum of it in a new variable `b`. For that, I would like you to explore two ways: using `if-else` block or using `Math.max` method. 🙂

Week1/homework/nadin.js
```diff
@@ -0,0 +1,111 @@
...
+
+let b = 8;
+console.log(b);
+
+let colors = [];
+console.log("the value of my variable y will be: an empty array");
+console.log(colors);
+
+let animals = ['dog', 'cat', 'horse', 'fish', 'bird'];
```
yash-kapila: You can use `const` here for both `colors` and `animals` variable since we aren't reassigning these two variables anywhere 🙂

Week1/homework/nadin.js
```diff
@@ -0,0 +1,111 @@
...
+console.log(animals);
+animals.push('baby pig');
```
yash-kapila, line 54: 👍  for using `push` method here

Week1/homework/nadin.js
```diff
@@ -0,0 +1,111 @@
...
+console.log("I think the type of my variable g is: null");
+console.log('d:string, e:number, f:boolean, g:null');
+
+console.log(typeof d);
+console.log(typeof e);
+console.log(typeof f);
+console.log(typeof g);
+
+
+console.log(e == d, f !== g);
```
yash-kapila: Hi Nadin, two things here please:
1. You should always use tripe equal operator `===` instead of `==` 
2. You are supposed to equate the `type` of a variable i.e `typeof e === typeof d` instead of the actual variable.

Could you please correct this and push your changes again? Please drop me a message on Slack if you have any questions. 🙂

Week1/homework/nadin.js
```diff
@@ -0,0 +1,111 @@
...
+
+
+// if (e = d ) {
+//    return "positive";
+// } else {
+//     return "not positive";
+// }
+
+
+if (e = d ) {
```
yash-kapila: Ah. You should be very careful here. Inside the if block, you are using a single equal operator i.e `=`. This is not an equality check but over here we are assigning value of `d` to variable `e`. This operation would always return a `truthy` value and hence would always go inside the `if` block.

You should never use `=` inside the if block but should be using `===`  operator. 🙂

Week1/homework/nadin.js
```diff
@@ -0,0 +1,111 @@
...
+
+let s = 10;
+console.log(s % 4);
+
+let p = 600;
+console.log(p % 7);
+
+let q = [15,'one'];
+
+//infinity is a numeric value representing infinity in javascript
```
yash-kapila, line 104: 👍

Week1/homework/nadin.js
```diff
@@ -0,0 +1,111 @@
...
+// }
+
+
+if (e = d ) {
+   console.log('SAME TYPE');
+}
+
+let m = 7;
+m = m % 3;
+console.log(m);
```
yash-kapila: What do you think the value would be of `m` here? 😉

yash-kapila (changes requested):
Hi @Nadoshka - I have reviewed your PR but I would like for you to go through the comments once and address the changes before I can approve them. Please feel free to drop me a message on Slack if you have any questions around those comments. 🙂

Week1/homework/nadin.js
Nadoshka: It's one :), of course I need to log more describtive message, thanks for your feedback!

Week1/homework/nadin.js
Nadoshka: Yes, sure! :) 👍

Week1/homework/nadin.js
Nadoshka: Thanks a lot! You are right, I mis-understood the requirement, that gave me the chance to read about Math.max , Math.min and I use it with if/if-else. ;) It's done!

Week1/homework/nadin.js
Nadoshka: It 's 7 , now I am sure! :) That helps me well! Thanks Yash👍

Week1/homework/nadin.js
Nadoshka: I will never forget it, Thanks 👍 :)

Week1/homework/nadin.js
Nadoshka: It is done! I sent u my q, Thanks for your help and patience!

wouterkleijn:
Dear @Nadoshka  Well-done! Thanks for sending in the homework in time and doing a great job.
Love the replies to Yash's comments. Very professional!

Nadoshka:
> Dear @Nadoshka Well-done! Thanks for sending in the homework in time and doing a great job.
> Love the replies to Yash's comments. Very professional!

Dear @wouterkleijn Thanks a lot for your kind feedback! That incourage me. I apprictiate it! :)

yash-kapila (approved):

PR #78, opened 2019-03-07
semih homework week2

sigma21:


semih.js
```diff
@@ -0,0 +1,119 @@
+/*Assignment #1*/
+/*printing hello world in a couple of languages with console.log function */
+console.log("Hello World // English");
+console.log("Merhaba Dünya // Turkish");
+console.log("Привет, мир // Russian");
+console.log("Hallo Wereld // Dutch");
```
yash-kapila, line 6: 👍

semih.js
```diff
@@ -0,0 +1,119 @@
...
+
+
+/*Assignment #2*/
+/*added \ for escape character*/ 
+console.log('I\'m awesome');
```
yash-kapila, line 11: Correct. Another way to do it would be to use double quotes `""` like this: `console.log("I'm awesome");` 🙂

semih.js
```diff
@@ -0,0 +1,119 @@
...
+console.log('I\'m awesome'); 
+
+
+/*Assignment #3*/
+let x;
+console.log("The value of my variable x will be: 'undefined'");
+console.log(x);
+x = 0;
+console.log("The value of my variable x will be: 0");
+console.log(x);
```
yash-kapila, line 20: 👍

semih.js
```diff
@@ -0,0 +1,119 @@
...
+
+/*Assignment #4*/
+let y = "some_string";
+console.log("The value of my variable y will be: 'some_string'");
+console.log(y);
+y = "new_string"
+console.log("The value of my variable y will be: 'new_string'");
+console.log(y);
```
yash-kapila, line 28: 👍

semih.js
```diff
@@ -0,0 +1,119 @@
...
+console.log(a);
+let highestvalue;
+
+if (a<z) {
+    highestvalue=z;
+} else {
+    highestvalue=a;
+}
+
+console.log(highestvalue);
```
yash-kapila, line 44: This is correct. However, I would like you to explore three things here:
1. You can use `const` keyword for both `z` and `a` variables instead of `let` since we aren't reassigning them to any new values.
2. Can you learn about the difference between `Math.floor` and `Math.round` methods 
3. You can also use `Math.max` method to select a maximum value from `a` and `z` instead of the `if-else` block

semih.js
```diff
@@ -0,0 +1,119 @@
...
+
+
+/*Assignment #6*/
+let items=[];
+console.log("The value of my array will be:'[]'");
+console.log(items);
+let favorite_animals=["horse", "lion", "tiger", "bird", "turtle"];
+console.log(favorite_animals);
+favorite_animals.push("baby pig");
```
yash-kapila, line 53: Even though this is absolutely correct functionally, the name of the variable doesn't fall in place with the naming convention concept which we learnt in Day 1. The variable names should follow the `camelCasing` convention instead of using underscores. This means, you should try to have your variable name as `favoriteAnimals` instead of `favorite_animals` 🙂

semih.js
```diff
@@ -0,0 +1,119 @@
...
+console.log(favorite_animals);
+
+
+/*Assignment #7*/
+let myString="this is a test";
+console.log(myString);
+let string_length=myString.length;
+console.log(string_length);
```
yash-kapila, line 61: Two things here please:
1. Variable name should follow camelCase as I have suggested above. I see you use the convention while naming `myString`. Wondering why you used underscore for the length variable.
2. You can use `const` keyword here instead of `let` to create these variables.

semih.js
```diff
@@ -0,0 +1,119 @@
...
+console.log("The value of the variable my_boolean is: " + my_boolean);
+console.log("The type of the variable my_boolean should be 'boolean'.");
+console.log(typeof(my_boolean));
+
+let my_array=[my_string,my_number,my_boolean];
+console.log("The value of the variable my_array is: " + my_array);
+console.log("The type of the variable my_array should be 'object'.");
+console.log(typeof(my_array));
+
+if (typeof(my_number) == typeof(my_string)) {
```
yash-kapila, line 85: A couple things here again:
1. Even though I am mentioning this the third time but please use camelCase for naming variables. It is very important to follow this convention as it makes you a better developer.
2. You should always use triple equals `===` operator instead of double equals `==` when trying to compare two values.

semih.js
```diff
@@ -0,0 +1,119 @@
...
+    console.log("different type");
+}
+
+
+/*Assignment #9*/
+console.log("% (modulo) gives us the remainder of a division.");
+console.log("so if we write the following statement: let x = 4%2; then the value of the variable x would be 0");
+console.log("here is an example:");
+let x=7;
+x=x%3;
```
yash-kapila, line 103: This is a very small thing but you should try to have a space between the operators like you do above i.e do `let x = 7;` instead of `let x=7;` as it is more readable.

semih.js
```diff
@@ -0,0 +1,119 @@
...
+console.log("If x=7, the answer of x=x%3 should be: ",x);
+
+
+/*Assignment #10*/
+let my_number=0;
+let my_string="some string";
+let my_boolean=true;
+let my_array=[my_string,my_number,my_boolean];
```
yash-kapila, line 111: Except the variables naming convention and use of `let` instead of `const`, everything else is correct.

yash-kapila (approved):
Hi @sigma21 - I have reviewed your PR and am approving the same. However, I would like you to go through the comments i.e use of `const` over `let` wherever possible and use `camelCase` instead of underscores while naming variables and please make sure that you try to address them in your next PR. 🙂

wouterkleijn:
Dear @sigma21  Well-done! Thanks for sending in the homework and getting your PR approved.

It would be nice if you can respond to @yash-kapila 's comments !

Keep up the good work.

semih.js
sigma21: I'll make sure I get used to using `const` over `let`.
Regarding the `Math.floor` & `Math.round`, I checked the MDN and as it is also stated there `Math.floor` returns the largest integer less than the given one. On the other hand, `Math.round` returns the nearest integer. So it could be higher or lower than the given one. 

example: `console.log(Math.floor(0.999))=0` and `console.log(Math.round(0.999))=1`

`Math.max` is also very useful and makes `if-else` unnecessary in our case as you mentioned. Thanks a lot for the additional info!

semih.js
sigma21: dulyNoted :)

semih.js
sigma21: Alright, I'll be extra careful about `const` and camelCase!

semih.js
sigma21: camelCase again :) Duly noted!
Regarding the triple equals, it is a fairly new concept to me, hence my lack of usage. I'll make sure I implement them in my code from now on.

semih.js
sigma21: I usually pay attention to the space between operators but apparently I missed these two. I'll be more careful.

semih.js
sigma21: Noted! `const` over `let`.

sigma21:
Dear @wouterkleijn  I didn't know that I can comment back here in this page. Thanks for the heads up!

sigma21:
Dear @yash-kapila I responded to your feedbacks on their respective windows. (right after Wouter's reprimand :) ) Thanks a lot for the top-notch evaluation. Appreciate the time you spend for this! :)

PR #79, opened 2019-03-07
Enes homework

eneserdek:


enes-homework.js
```diff
@@ -0,0 +1,145 @@
+//Q1
+
+const Turkish = 'Merhaba Dunya',
+  English = 'Hello, World',
+  French = 'Bonjour, le Monde',
+  Dutch = 'Hoi, de wereld';
+console.log(Turkish, English, French, Dutch);
```
yash-kapila, line 7: I liked that you used `const` here 🙂

enes-homework.js
```diff
@@ -0,0 +1,145 @@
...
+
+//Q2
+console.log('I'm awesome');
```
yash-kapila: Ah. You need to comment out this line as well otherwise the program throws an error otherwise. 🙂

enes-homework.js
```diff
@@ -0,0 +1,145 @@
...
+console.log('I\'m awesome');
+console.log("I'm awesome");
+
+//Q3
+let x;
+console.log("the value of x will be : undefined");
+console.log(x);
+x = 7;
+console.log("the value of x will be : 7");
+console.log(x);
```
yash-kapila, line 21: 👍

enes-homework.js
```diff
@@ -0,0 +1,145 @@
...
+
+//Q4
+const y = "I like birds";
+console.log("the value of y will be : I like birds");
+console.log(y);
+y = "I hate birds";
+console.log("the value of y will be : error because I have used const as a variable. The case will be different if I had used var or let. While debugging it does not throw an error message I don't know why.");
+console.log(y);
```
yash-kapila, line 29: Hi Enes - you are right here that it would throw an error while using `const` keyword. It would work just fine if you used `let` instead. However, I am not sure why you say that you didn't see any error while using `const`. I am quite certain that you should see an error like this: `Uncaught TypeError: Assignment to constant variable.`. Can you please check again?

enes-homework.js
```diff
@@ -0,0 +1,145 @@
...
+let a = Math.round(z);
+console.log(a);
+
+//The rest is same;
+if (z - a > 0) {
+  console.log(z)
+}
+else {
+  console.log(a)
+}
```
yash-kapila, line 56: Hi Enes - I am not certain if I understood your logic in **Method 1**. Could you please elaborate it a little for me?

The method 2 is correct. However, I would like if you tried to search for the difference between `Math.floor` vs `Math.round` methods. Also, for the last part of the assignment, you need to assign the higher of `z` and `a` to a new variable say `max`. How would you do it? 🙂

enes-homework.js
```diff
@@ -0,0 +1,145 @@
...
+
+//Q6
+
+const emptyArrOfAnimals = [];
+console.log("emptyArrOfAnimals : empty array");
+console.log(emptyArrOfAnimals);
+
+const myFavouriteAnimals = ["Budgie Parrots", "German Shephard", "Goat"];
+console.log(myFavouriteAnimals);
+myFavouriteAnimals[3] = "baby pig"; //myFavouriteAnimals.push("baby pig") method can also be used
```
yash-kapila: That's correct. I would suggest that you use the `push` method instead of doing it this way 😉

enes-homework.js
```diff
@@ -0,0 +1,145 @@
...
+console.log(myFavouriteAnimals);
+
+//Q7
+
+let myString = "this is a test";
```
yash-kapila, line 73: It can be created using `const`, right?

enes-homework.js
```diff
@@ -0,0 +1,145 @@
...
+console.log("typeof firstVar:number; typeof secondVar: string; typeof thirdVar: booelan; typeof fourthVar: string")
+
+typeof firstVar;
+typeof secondVar;
+typeof thirdVar;
+typeof fourthVar;
+
+typeof firstVar === typeof secondVar;
+typeof thirdVar !== typeof fourthVar;
+typeof secondVar === typeof fourthVar;
```
yash-kapila: If you remember from our class today when we learnt about statements and expressions, lines 90 to 92 are expressions which evaluate into a `true` or a `false` value. However, we don't do anything with these values. It would make more sense if you perhaps `console.log` these values. 🙂

yash-kapila (changes requested):
Hi @eneserdek - I have reviewed your PR but before approving the same, I would like you to please address the comments in assignment **#5**. Could you please have a look? 🙂

enes-homework.js
eneserdek: //console.log(I'm awesome') is omitted.

enes-homework.js
eneserdek: when I redo, it throw out an error as expected.

enes-homework.js
eneserdek: My logic for rounding the number is like this : 
7.25 - 0.50 = 6.75. 6.75 is smaller than 7. So, 7.25 would be rounded to 7. if it was 7.51 than 7.51 - 0.5 = 7.01. This is bigger than 7. So in this case 7.51, would be rounded to 8.

the code to assign biggest one to variable max:
let max = (z>a) ? z : a;

wouterkleijn:
Dear @eneserdek Well-done! Congrats on getting your PR approved and we appreciate your direct response to Yash's comments. 

Keep up the good work.

PR #80, opened 2019-03-07
JS1 Homework Week 2 - Wadeea

WadeeaKiwan:


Week2/wadeea.js
```diff
@@ -0,0 +1,246 @@
+// 1. Arabic - English - Dutch - Italian -Spanish - French.
+console.log(
+  `مرحبا أيها العالم
+  Hello, World!
+  Hallo, wereld!
+  Ciao, mondo!
+  Hola, mundo!
+  Bonjour, monde`,
+);
```
yash-kapila, line 9: 👍

Week2/wadeea.js
```diff
@@ -0,0 +1,246 @@
...
+
+// 2. The error was because of using the same quotes as the apostrophe in 'I'm'.
+console.log(`I'm awesome`);
```
yash-kapila, line 12: That's correct. You can also use double quotes like: `console.log("I'm awesome");` or use escape characters i.e backslash `\` like: `console.log('I\'m awesome');`. This is just for your information.

Week2/wadeea.js
```diff
@@ -0,0 +1,246 @@
...
+console.log(`The value of my variable x will be: undefined`);
+// 3.3
+console.log(x);
+// 3.4 If I initialize the variable x here, I will get a syntax error,
+// because It's already been declared. So I've only defined it.
+x = 7;
+// 3.5
+console.log(`The value of x will be: 7`);
+// 3.6
+console.log(x);
```
yash-kapila, line 32: 👍

Week2/wadeea.js
```diff
@@ -0,0 +1,246 @@
...
+// 4.1
+console.log(`The value of the string will be the following sentence: This is a string`);
+// 4.2
+console.log(y);
+// 4.3
+y = `New string`;
+// 4.4
+console.log(`The value will be the following: New string`);
+// 4.5
+console.log(y);
```
yash-kapila, line 45: 👍

Week2/wadeea.js
```diff
@@ -0,0 +1,246 @@
...
+let a = 7;
+// 5.4
+console.log(a);
+// 5.5
+let theHighestValue;
+if (a > z) {
+  theHighestValue = a;
+} else {
+  theHighestValue = z;
+}
```
yash-kapila: Hi Wadeea, a few points to note here please:
1. You can use `const` keyword here instead of `let` for declaring `z` and `a`.
2. You need not hard-code the value for variable `a`. Please try to explore `Math.round` and `Math.floor` methods. 🙂 
3. For finding the higher value between `a` and `z`, the `if-else` works just fine. However, I would like you to explore `Math.max` method here.

Week2/wadeea.js
```diff
@@ -0,0 +1,246 @@
...
+// 5.6
+console.log(`The highest value is: ` + theHighestValue);
+
+// 6.1
+const citiesInNetherlands = [];
+// 6.2
+console.log(`The value of the array is: 0`);
```
yash-kapila: Is the value of this array `0` or an empty array `[]`? 😉

Week2/wadeea.js
```diff
@@ -0,0 +1,246 @@
...
+// 6.3
+console.log(citiesInNetherlands);
+// 6.4
+const favoriteAnimals = [`cat`, `dog`, `hamster`, `lion`, `tiger`];
+// 6.5
+console.log(favoriteAnimals);
+// 6.6
+favoriteAnimals.push(`baby pig`);
```
yash-kapila, line 72: 👍

Week2/wadeea.js
```diff
@@ -0,0 +1,246 @@
...
+// 6.7
+console.log(favoriteAnimals);
+
+// 7.1
+const myString = `this is a test`;
+console.log(myString);
+// 7.2
+myString.length;
```
yash-kapila: You need not have this in the code. If you remember from our class today i.e difference between `statements` and `expressions`, line 79 `myString.length` is an expression which isn't really understood by the computer as we don't do anything with the evaluated value of this expression. This makes more sense in line 81 where you actually use the value of the expression to print it on the console.

Week2/wadeea.js
```diff
@@ -0,0 +1,246 @@
...
+// 7.3
+console.log(myString.length);
+
+// 8.1
+const myFirstName = `Wadeea`;
+const myLastName = `Kiwan`;
+let myAge = 33;
```
yash-kapila: Why did you decide to go with `let` for `age`?

Week2/wadeea.js
```diff
@@ -0,0 +1,246 @@
...
+typeof myLastName === typeof myAge;
+typeof myLastName === typeof myTelephone;
+typeof myLastName === typeof genderMale;
+typeof myLastName === typeof childrenFirstNames;
+typeof myAge === typeof myTelephone;
+typeof myAge === typeof genderMale;
+typeof myAge === typeof childrenFirstNames;
+typeof myTelephone === typeof genderMale;
+typeof myTelephone === typeof childrenFirstNames;
+typeof genderMale === typeof childrenFirstNames;
```
yash-kapila: Again, these are simply expressions which don't really add much value to our program as we don't use them anywhere. It would be better if you maybe printed these values as `console.log(typeof myFirstName === typeof myLastName)`. I liked how you followed the naming conventions while creating your variables. Thumbs up for that.

Week2/wadeea.js
```diff
@@ -0,0 +1,246 @@
...
+// 8.6
+if (typeof myFirstName === typeof myLastName) {
+  console.log(`SAME TYPE`);
+} else {
+  console.log(`NOT SAME TYPE`);
+}
+
+if (typeof myFirstName !== typeof myAge) {
```
yash-kapila: Should it not be `===` instead of `!==` here?

Week2/wadeea.js
```diff
@@ -0,0 +1,246 @@
...
+m %= 4;
+console.log(`the new value of m will be: 0`);
+console.log(m);
+let n = 19;
+n %= 5;
+console.log(`the new value of n will be: 4`);
+console.log(n);
+let p = 6;
+p %= 7;
+console.log(`the new value of p will be: 6`);
```
yash-kapila, line 234: 👍

Week2/wadeea.js
```diff
@@ -0,0 +1,246 @@
...
+
+// 10.1
+const myProfile = [`Wadeea`, `Kiwan`, 33, 0612345678];
+console.log(typeof myProfile[0]);
+console.log(typeof myProfile[2]);
+// 10.2 - 10.3
+6 / 0 === 10 / 0;
+console.log(`The comparison result will be: true`);
+console.log(6 / 0 === 10 / 0);
+console.log(100 / 0 === 15 / 0);
```
yash-kapila, line 245: 👍

yash-kapila (approved):
Hi @WadeeaKiwan - I have reviewed your PR and am approving the same. However, could you please go through the comments I have added? Please drop me a message on Slack if you have any queries around them.

Week2/wadeea.js
WadeeaKiwan: Thanks. I'll consider this one also. I couldn't use backslash because VSCode removes it every time I save the file.

Week2/wadeea.js
WadeeaKiwan: Thanks. Of course it is easier to use these methods. I didn't know them before. I've just explored them. 🙂

Week2/wadeea.js
WadeeaKiwan: Yes, of course an empty array []. 😉

Week2/wadeea.js
WadeeaKiwan: I'll correct this one also.😉

Week2/wadeea.js
WadeeaKiwan: It was maybe from the last week that we wanted to manipulate it, but it's not the case here. I'll change it to const.

Week2/wadeea.js
WadeeaKiwan: Thanks a lot 🙂. I'll print them.

Week2/wadeea.js
WadeeaKiwan: Yes, you're right.

WadeeaKiwan:
Hi Yash, @yash-kapila 
I've just corrected the points you mentioned. I pushed the last version. Do I have to make a new pull request? And how? Because I can see the changes on github, but I couldn't make pull request.
Many thanks for your feedback and your time.🙂

yash-kapila:
Hi @WadeeaKiwan - when you address any changes requested in your current pull request and push to the same branch, you need not create a new pull request. This is because all your changes would be visible in the existing PR.

wouterkleijn:
Dear @WadeeaKiwan Well-done! Thanks for sending in the homework in time and doing a great job.

We also appreciate that you respond to Yash's comments

Keep up the good work.

PR #82, opened 2019-03-08
JS1 Week2

saidcetinkaya:


saidcetinkaya.js
```diff
@@ -0,0 +1,133 @@
+// Question 1 //
+
+console.log('Hello World')
+console.log('Halo Dunia ')
+console.log('Hola Mundo')
```
yash-kapila, line 5: Ah. You missed putting semi-colons here.

saidcetinkaya.js
```diff
@@ -0,0 +1,133 @@
...
+
+
+
+// Question 2 //
+
+console.log('I\'m awesome');
+console.log("I'm awesome");
```
yash-kapila, line 12: 👍

saidcetinkaya.js
```diff
@@ -0,0 +1,133 @@
...
+
+
+// Question 3 //
+
+let x;
+console.log('the value of my variable x will be: undefined');
+console.log(x);
+x = 1;
+console.log("the value of my variable x will be: 1");
+console.log(x);
```
yash-kapila, line 23: 👍

saidcetinkaya.js
```diff
@@ -0,0 +1,133 @@
...
+
+
+// Question 4 //
+
+let y = 'string';
+console.log('y will be: string');
+console.log(y);
+y = 'not the one before';
+console.log('y will be: not the one before');
+console.log(y);
```
yash-kapila, line 34: 👍

saidcetinkaya.js
```diff
@@ -0,0 +1,133 @@
...
+
+
+// Question 5 //
+
+let z = 7.25;
+console.log(z);
+let a = Math.round(z);
+console.log(a);
+let highest = Math.max(a, z);
+console.log(highest);
```
yash-kapila, line 45: I would like you to explore two things here please:
1. What's the difference between `Math.round` and `Math.floor`.
2. You can also make use of `const` keyword here to create `z`, `a` and `highest` variables instead of `let` since we aren't reassigning them anywhere, right? 🙂

saidcetinkaya.js
```diff
@@ -0,0 +1,133 @@
...
+
+
+// Question 6 //
+
+let wildOnes = [];
+console.log('wildOnes will be: []');
+console.log(wildOnes);
+let animals = ['snake', 'cat', 'dog', 'horse'];
+console.log(animals);
+animals.push('baby pig');
```
yash-kapila, line 56: 👍

saidcetinkaya.js
```diff
@@ -0,0 +1,133 @@
...
+console.log(animals);
+
+
+
+// Question 7 //
+
+let myString = "this is a test";
```
yash-kapila, line 63: Can use `const` here, right? 🙂

saidcetinkaya.js
```diff
@@ -0,0 +1,133 @@
...
+// Question 8 //
+
+let b = 'string', c = 1, d = ['array'], e = {'key':'value'};
+console.log("The value of my variable a is: " + b);
+console.log("The value of my variable b is: " + c);
+console.log("The value of my variable c is: " + d);
+console.log("The value of my variable d is: " + e);
+console.log('b is string');
+console.log('c is number');
+console.log('d is array');
```
yash-kapila, line 78: Are you sure `c`'s type is an `array` and not an `object`? Did you see anything weird? 😉

saidcetinkaya.js
```diff
@@ -0,0 +1,133 @@
...
+    for(j = i + 1; j < Object.keys(controlObject).length; j++) {
+        if (typeof(Object.values(controlObject)[i]) == typeof(Object.values(controlObject)[j])){
+            console.log(Object.keys(controlObject)[i] + ' and ' + Object.keys(controlObject)[j] + ' are same type');
+        }
+        else {
+            console.log(Object.keys(controlObject)[i] + ' and ' + Object.keys(controlObject)[j] + ' are not same type');
+        }
+    }
+}
+
```
yash-kapila, line 101: Hi Said, couple things here for your information:
1. `controlObject` can be `const` instead of `let`.
2. At the moment it is fine, but as you proceed further and get more comfortable with working with objects, you wouldn't need to explicitly enclose the object properties in single quotes as they are implicitly considered as strings i.e you can simply do like:
```
const controlObject = {
  b: b,
  c: c,
  d: d,
  e: e
};
```
3. You need not declare `i` and `j` explicitly on line 90. You can include them inside the `for` loop as you would have noticed we did during our class on Sunday.
4. You should always use triple equals `===` instead of double equals `==` when doing a comparison. 🙂

yash-kapila (approved):
Hi @saidcetinkaya - I have reviewed your PR and am approving the same. However, I do have put in some comments. Can you please have a look at them and make sure that you incorporate them in your future homework submissions? Have a nice day ahead. 🙂

wouterkleijn:
Dear @saidcetinkaya Great job! We appreciate your efforts.

It would be nice if you can respond to and address some of @yash-kapila 's comments !

Keep up the good work.

saidcetinkaya.js
saidcetinkaya: Oppps.
I'll try harder not to miss it

saidcetinkaya.js
saidcetinkaya: 1. Math.floor is rounding to downward. i.e: Math.floor(5.95) // output:5. 
    However Math.round(5.95) = 5.
2. I just tend to use 'let' more than 'const'. But you are right, we aren't reassigning them anywhere.

saidcetinkaya.js
saidcetinkaya: 👍

saidcetinkaya.js
saidcetinkaya: 👍

PR #85, opened 2019-03-13
Enes homework week3

eneserdek:


enes-homework.js
```diff
@@ -0,0 +1,145 @@
+//Q1
```
yash-kapila, line 1: Hi Enes, there is one small issue with this homework. I think you created a new branch for week 3's homework from the week 2's homework branch instead of doing it from `master`. Thus, I can see your week2 homework as well along with week3. Could you please take care of it the next time onwards to always create a new branch for your homework from the `master` branch. 🙂

enes-homework-week3-strings -and-arrays.js
```diff
@@ -0,0 +1,36 @@
+//Q1
+
+let myString = 'hello,this,is,a,difficult,to,read,sentence';
+console.log("myString: the result will be 'hello,this,is,a,difficult,to,read,sentence'");
+console.log(myString);
+console.log('the lenght of myString will be 42');
+console.log(myString.length);
+
+let myNewString = myString.split(',').join(' ');
```
yash-kapila, line 9: 👍 

I've a couple things to add here though:
1. You can avoid creating a new variable `myNewString` here and can simply do `myString = myString.split(',').join(' ');`
2. This approach of doing `myString.split(',').join(' ');` is called **method chaining**. You will probably see this a lot as you move forward with JavaScript.

enes-homework-week3-strings -and-arrays.js
```diff
@@ -0,0 +1,36 @@
...
+//Q2
+
+let favoriteAnimals = ['blowfish', 'capricorn', 'giraffe'];
+favoriteAnimals.push('turtle');
+console.log('favoriteAnimals : the result will be: blowfish, capricorn, giraffe, turtle');
+console.log(favoriteAnimals);
+
+favoriteAnimals.splice(1, 0, 'meerkat');
+console.log(favoriteAnimals); //expected output: blowfish meerkat capricorn giraffe turtle
+console.log(favoriteAnimals.length); //expected output: 5
```
yash-kapila, line 25: This is correct but you will learn a little later that using `splice` actually changes the `favoriteAnimals` array itself. This is called mutating the data structure and often it is suggested that we don't mutate the data structure i.e try to push the `meerkat` in between the two elements but return a new array itself thus having two arrays at the end, original and the modified array.

Maybe when you are introduced to this concept, you can revisit it and think of a way by which you can still add `meerkat` but don't modify the existing array. 🙂

enes-homework-week3-moreJS.js
```diff
@@ -0,0 +1,200 @@
+//Q:More JS
+
+//Q1
+
+function sum(a,b,c) {
+  return (a+b+c);
```
yash-kapila, line 6: This is called as a pure function where the function works on the inputs provided to it as arguments i.e `a`, `b` and `c` and then returns a computed value based on those 3 arguments.

enes-homework-week3-moreJS.js
```diff
@@ -0,0 +1,200 @@
...
+};
+
+console.log(sum(1,2,3));
+
+//Q2
+
+function colorCar(str) {
+  console.log("a " + str + " car" );
```
yash-kapila, line 14: Aahaan. Can you think of an alternative for double quotes here? Perhaps using template strings(backticks) 😉

enes-homework-week3-moreJS.js
```diff
@@ -0,0 +1,200 @@
...
+//Q3
+const obj1 = {
+  name: "Jason",
+  surname: "Born",
+  age: 33,
+};
+
+function printOutPropertiesAndValues (obj) {
+  for (let key in obj) {
+    console.log(key);
```
yash-kapila, line 31: Hi Enes, this function works fine. However, if you remember from our class, I mentioned that you can try to use `Object.keys` instead of a `for-in` loop. You can perhaps try to google the difference between the two approaches and see if it makes sense. Otherwise, I think this is going to be covered in JS3. 🙂

enes-homework-week3-moreJS.js
```diff
@@ -0,0 +1,200 @@
...
+    when selected other than 1,2 with this method.)
+  function vehicleType(color, code) {
+    vehicles = (code ===1 ) ? "car" :
+              (code === 2 ) ? "motorcyle" : 
+              "";
+  console.log("a " + color + " " + vehicles);
+  }
+  */
+
+vehicleType("blue", 1);
```
yash-kapila, line 69: Hi Enes, couple things here:
1. I understand that you wanted to split out your code into two functions and I appreciate your thinking. However, I think the final solution is a bit complicated and could have been easier. Downfall of the approach is that you have to now create a sort-of global variable i.e `let vehicles = "";` which you then manipulate in `codeDesigner` function and it is then read inside the `vehicleType` function. I think I would have kept a single function for simplicity sake and inside the `if-elseif-else` had the console log statements thus avoiding the need of a new variable `vehicles`. 

2. Using conditional operators in this case isn't the best of ideas because of the reason which you yourself have mentioned in the comments. Maybe try using a `switch` block instead 🙂

enes-homework-week3-moreJS.js
```diff
@@ -0,0 +1,200 @@
...
+
+
+//Q5
+
+console.log((3===3) ? "yes" : "no"); //Since .log is an function it is possible to write down an expression in it.
```
yash-kapila, line 74: 👍

enes-homework-week3-moreJS.js
```diff
@@ -0,0 +1,200 @@
...
+}
+function vehicleType(color, code) {
+  codeDesigner(color, code);
+  console.log("a " + color + " " + vehicles);
+}
+*/
+let situation = "";
+function vehicle(color, code, age){
+  codeDesigner(color, code);
+  situation = (age == 0) ? "brand-new" : "used";
```
yash-kapila, line 97: `==` or `===`? 😉

yash-kapila: And how about using the back-ticks in console log?

enes-homework-week3-moreJS.js
```diff
@@ -0,0 +1,200 @@
...
+console.log(listOfVehicles[2]); // expected output: bike
+
+//Q9 
+let vehicles = "";
+function codeDesigner(color, code){
+  vehicles = type = (code === 1) ? "motorbike" : 
+         (code === 2 ) ? "caravan" :
+         (code === 3 ) ? "bike" :
+         (code === 4 ) ? "plane" :
+         (code === 5 ) ? "yacht" : "";
```
yash-kapila, line 119: I think you perhaps mis-understood the requirement. You were supposed to use the list `listOfVehicles` instead of using a conditional operator. Something like:
```
vehicles = listOfVehicles[code];
```

Also, I am not sure what `type` is here. Could you please explain line 115 to me?

enes-homework-week3-moreJS.js
```diff
@@ -0,0 +1,200 @@
...
+}
+
+
+vehicle("green", 3, 1);
+
+//Q10
+
+let joesGarage = ["car", "motorbike", "caravan", "bike", "plane", "yacht"];
+let arr;
```
yash-kapila, line 128: Do we need this `arr` variable?

enes-homework-week3-moreJS.js
```diff
@@ -0,0 +1,200 @@
...
+  console.log("Amazing Joe's Garage, we service " + arr2 + ".");
+}
+
+advertisementMaker(joesGarage);
+
+//Q11
+
+joesGarage.push("tank");
+advertisementMaker(joesGarage);
+//normally it would work. but somehow, the functions works in the question 10 but not in question 11. I have solved this problem buy defining an array which is valid only inside the function.
```
yash-kapila, line 145: Hi Enes, could you please elaborate on this a bit more? What do you mean that the function doesn't work for Q11 if you declared `arr2` inside the function?

enes-homework-week3-moreJS.js
```diff
@@ -0,0 +1,200 @@
...
+
+//Q13
+const teacherNames = {
+  HTML_CSS : ["Philip", "Rob"];
+  CLI : ["Unmesh", "Hardith"];
+  JS1 : ["Yash"];
+};
+
+//Q14
+teacherNames.languages = ["English"];
```
yash-kapila, line 158: Ah I think you mis-understood the requirement again but looks like you did it anyways in Q13. 🤣 

Languages meant were `HTML/CSS`, `CLI` and `JS1`. I think the wording in the homework could have been better. Perhaps `topics` instead of `languages` would be better suited

enes-homework-week3-moreJS.js
```diff
@@ -0,0 +1,200 @@
...
+    }
+  }
+
+  x == y // expected output: false;
+  x === y // expected output: false;
+  z == y // expected output : true;
+  z === y // expected output: true
+
+  equalityTester(x, y); /*expected output: x and y are not equal but z and y are equal. Even though the elements of x and y are the same, 
+  because they are referencing to different array data's, java does not seem them as equals.
```
yash-kapila, line 181: `java` or `javascript`? 😉

enes-homework-week3-moreJS.js
```diff
@@ -0,0 +1,200 @@
...
+      console.log(a + " and " + b + "are equal but they may not be strictly equal");
+    } else {
+      console.log(a + " and " + b + " are not equal");
+    }
+  }
+
+  x == y // expected output: false;
+  x === y // expected output: false;
+  z == y // expected output : true;
+  z === y // expected output: true
```
yash-kapila, line 178: 👍

enes-homework-week3-moreJS.js
```diff
@@ -0,0 +1,200 @@
...
+
+//Q16.
+let o1 = { foo: "bar" };
+let o2 = { foo: "bar" };
+let o3 = o2;
+
+o2.name = "camel";
+console.log(o2); // expected output: foo: "bar", name: "camel";
+console.log(o3); // expected output: foo: "bar", name: "camel";
+console.log(o2 === o3); // exoected output: true;
```
yash-kapila, line 192: 👍

yash-kapila (commented):
Hi Enes, good evening. Apologies first for leaving this so late but I have now finished reviewing your assignment. Even though there aren't any significant changes needed, I would like you to go through the comments and perhaps see to it that those are addressed in your future PRs. However, I do have a question which I have put for Question 9. Could you please check it once?

enes-homework.js
eneserdek: I did not know that if opening a new branch from someother branch would be different than opening a branch from master. I will keep your advise.

enes-homework-week3-strings -and-arrays.js
eneserdek: I have found the solution. But i still don't know what type of problems/bugs that a mutation can do.
here is the solution:  
const newArr = favoriteAnimals.splice(0,2).concat("meerkat").concat(favoriteAnimals(2,6); is this true?

enes-homework-week3-moreJS.js
eneserdek: wow thank you for reminding that. It is really better to use backticks. console.log(`a ${str} car`);

enes-homework-week3-moreJS.js
eneserdek: When I declared arr2 outside of the function advertisementMaker(arr), the function works only for once. When you add new item to the array on the next line and recall function again, the function would not work.  Because arr2 points some data in it because of the first action. Then I put this arr2 inside the function the problem resolved.

enes-homework-week3-moreJS.js
eneserdek: Thanks again. I made some readings and learnt to some extent the difference.

enes-homework-week3-moreJS.js
eneserdek: I could have created this variable inside the function. I divided the function in two because I wanted to use the function in the coming questions.

enes-homework-week3-moreJS.js
eneserdek: surely ===. and console.log(`a ${color} ${situation} ${vehicles});

enes-homework-week3-moreJS.js
eneserdek: type is a variable that I created whilst working on the question. I forgot to delete it. Secondly, vehicles = listOfVehicles[code] is a way better solution. Mine solution was a little bit hardcoding which is not desirable.

enes-homework-week3-strings -and-arrays.js
yash-kapila: Ah. Not quite there yet. Because you are still using `splice` method which would mutate the `favoriteAnimals` array again. Hint: try exploring `slice` method instead and see if you can somehow make it work that way.

enes-homework-week3-moreJS.js
yash-kapila: Yes. That is because earlier `arr2` was a sort-of global variable and hence when we modified it inside the function, it affected all instances of the variable. Which is why I suggested in our class to keep your functions pure(don't update any variable outside the function's ownership) and this is something you will be hearing often. 🙂

yash-kapila (approved):

PR #87, opened 2019-03-13
Akif homework week3

Akif2709:


Week2/akif.js
```diff
@@ -0,0 +1,108 @@
+// Part 1 Hello World
```
yash-kapila, line 1: This file shouldn't be here from Week2, right? Make sure for your future modules to always switch to `master` branch before creating a new branch for a new week's homework. 🙂

Week3/akif-week3.js
```diff
@@ -0,0 +1,158 @@
+//   STRINGS! --------------
+let myString = 'hello,this,is,a,difficult,to,read,sentence';
+console.log(myString);
+console.log(myString.length);
+myString = myString.replace(/[,]/g, ' ');
```
yash-kapila, line 5: Quite nice 👍

Week3/akif-week3.js
```diff
@@ -0,0 +1,158 @@
...
+favoriteAnimals.push('turtle');
+console.log(favoriteAnimals);
+favoriteAnimals.splice(1, 0, 'meerkat');
+console.log("i added a new value 'meerkat' to the array");
+console.log(favoriteAnimals);
+console.log('The array has a length of: ' + favoriteAnimals.length);
+favoriteAnimals.splice(3, 1);
+console.log(favoriteAnimals);
+console.log('The item you are looking for is at index: ' + favoriteAnimals.indexOf('meerkat'));
+favoriteAnimals.splice(favoriteAnimals.indexOf('meerkat'), 1);
```
yash-kapila, line 19: This is correct but you will learn a little later that using splice actually changes the favoriteAnimals array itself. This is called mutating the data structure and often it is suggested that we don't mutate the data structure i.e try to push the meerkat in between the two elements but return a new array itself thus having two arrays at the end, original and the modified array.

Maybe when you are introduced to this concept, you can revisit it and think of a way by which you can still add meerkat but don't modify the existing array. 🙂

Week3/akif-week3.js
```diff
@@ -0,0 +1,158 @@
...
+
+//  MORE JAVASCRIPT ! ------------------------------------------------
+// -----1111111111111111111-----------
+function getSum(a, b, c) {
+  return a + b + c;
```
yash-kapila, line 24: A pure function this is since it is not updating any variable outside its ownership 🙂

Week3/akif-week3.js
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// -----222222222222222222-------------
+function colorCar(color) {
+  return 'a ' + color + ' car';
```
yash-kapila: Can you think of an alternative here? Perhaps using template strings i.e backticks. 😉

Week3/akif-week3.js
```diff
@@ -0,0 +1,158 @@
...
+}
+
+//-----3333333333333333333--------------
+const myBooks = {
+  Dostoyevsky: 'Crime and Punishment',
+  'Franz Kafka': 'The Trial',
+  'Anne Frank': 'Diarys',
```
yash-kapila: Spelling mistake in the book name 🙂

Week3/akif-week3.js
```diff
@@ -0,0 +1,158 @@
...
+};
+
+function getBook(library) {
+  for (let book in library) {
+    console.log(book + ' - ' + library[book]);
```
yash-kapila, line 41: Hi Akif this function works fine. However, if you remember from our class, I mentioned that you can try to use `Object.keys` instead of a `for-in` loop. You can perhaps try to google the difference between the two approaches and see if it makes sense. Otherwise, I think this is going to be covered in JS3. 🙂

Week3/akif-week3.js
```diff
@@ -0,0 +1,158 @@
...
+//  _-------555555555555555555--------------------
+console.log(3 === 3 ? 'yes' : 'no');
+
+//  -------66666666666666666666---------------------
+function vehicle(color, code, age) {
+  if (age > 0) {
+    age = 'used';
+  } else {
+    age = 'new';
+  }
```
yash-kapila: Umm... I would maybe not overwrite `age` with a new value and instead do something like `const type = (age > 0) ? 'used' : 'new';` 🙂

Week3/akif-week3.js
```diff
@@ -0,0 +1,158 @@
...
+//   }
+
+//   arrname[arrname.length - 2] += ' and';
+
+//   console.log("Amazing Joe's Garage, we service " + arrname.join(' ') + '.');
+// }
+
+function advertisement(arrname) {
+  let vehicles1 = arrname.slice(0, arrname.length - 1).join('s, ');
+  let lastVehicle = arrname[arrname.length - 1];
```
yash-kapila: `let` or `const` here? 🙂

Week3/akif-week3.js
```diff
@@ -0,0 +1,158 @@
...
+vehicleList.push('ship');
+advertisement(vehicleList);
+
+// ---12 12 12 12 12 12 12 12 // 13 13 13 // 14 14 14 14 -------------------------
+const teachers = {};
+teachers.HTML = 'Philipp Beau';
+teachers.CSS = 'Rob van Kruijsdijk';
+teachers.GIT = 'Unmesh Joshi';
+teachers.JS = 'Yash Kapila';
+console.log(teachers);
```
yash-kapila, line 121: Hi Akif, is there a reason why you didn't put all the values during the object declaration itself i.e
```
const teachers = {
  HTML: 'Philipp Beau',
  CSS: 'Rob van Kruijsdijk',
  GIT: 'Unmesh Joshi',
  JS: 'Yash Kapila'
};
```

Week3/akif-week3.js
```diff
@@ -0,0 +1,158 @@
...
+let x = [1, 2, 3];
+let y = [1, 2, 3];
+let z = y;
+
+console.log(x == y); // false
+console.log(x === y); // false
+console.log(x == z); // false
+console.log(x === z); // false
+console.log(y == z); // true
+console.log(y === z); // true
```
yash-kapila, line 133: 👍

Week3/akif-week3.js
```diff
@@ -0,0 +1,158 @@
...
+
+// When i use == or ===, it compares whether they are same array or not.
+// to look they have a same content:
+console.log(x.toString() === y.toString()); // true
```
yash-kapila, line 137: Correct. This is because `toString` method on an array converts all its contents to string values and since both arrays have same elements, this results in a `true` value.

Week3/akif-week3.js
```diff
@@ -0,0 +1,158 @@
...
+o2.foo = 'o2-changed';
+console.log(o3); //changing o2 change o3
+o1.foo = 'o1-changed';
+console.log(o3); //changing o1 doesnt change o3
+// **Does the order that you assign (o3 = o2 or o2 = o3) matter?
+// Yes absolutely matter. if we write o2 = o3, o2 would be declared twice and cause ERROR.
+
+// ----17 17 17 17 17 17 17 17 ----------------
+let bar = 42;
+console.log(typeof typeof bar); // string
```
yash-kapila, line 153: 👍

Week3/akif-week3.js
```diff
@@ -0,0 +1,158 @@
...
+let o1 = { foo: 'bar' };
+let o2 = { foo: 'bar' };
+let o3 = o2;
+
+o2.foo = 'o2-changed';
+console.log(o3); //changing o2 change o3
+o1.foo = 'o1-changed';
+console.log(o3); //changing o1 doesnt change o3
+// **Does the order that you assign (o3 = o2 or o2 = o3) matter?
+// Yes absolutely matter. if we write o2 = o3, o2 would be declared twice and cause ERROR.
```
yash-kapila, line 149: 👍

yash-kapila (approved):
Hi @Akif2709 - Apologies for leaving it so late but I have reviewed the PR and am approving the same. However, I would suggest that you go through the comments once please and see if you can explore things asked in them. Have a good day ahead.

Week3/akif-week3.js
Akif2709: i know to do this but i think the directives in the homework was not clear. 
"12.Create an empty object.
13.Create an object that contains the teachers that you have had so far for the different modules.
14.Add a property to the object you just created that contains the languages that they have taught you." 😃

Week3/akif-week3.js
Akif2709: Hi Yash. You mean something like  this ?
let a = favoriteAnimals.slice(1, favoriteAnimals.length);
let b = favoriteAnimals.slice(0, 1);
b.push('meerkat');
b.push(...a);

Week3/akif-week3.js
Akif2709: Thanks for remarking subtle issues like that. Also for other comments. I think your comments is very useful 👍

Week3/akif-week3.js
yash-kapila: Yes. Something similar 🙂

PR #88, opened 2019-03-13
Orhan homework week3

OrhanGoc:


orhan-homework-week3.js
```diff
@@ -0,0 +1,167 @@
+//1. Part Strings!
+
+let myString = "hello,this,is,a,difficult,to,read,sentence";
+console.log(myString);
+console.log(myString.length);
+console.log(myString.replace(/,/g, " "));
```
yash-kapila, line 6: 👍 for using regex but 👎 for not using `const` 🙂

orhan-homework-week3.js
```diff
@@ -0,0 +1,167 @@
...
+let favoriteAnimals = ["blowfish", "capricorn", "giraffe"];
+favoriteAnimals.push("turtle");
+console.log(favoriteAnimals);
+favoriteAnimals.splice(1, 0, "meerkat");
+console.log("The new value of the array will be: " + "blowfish, meerkat, capricorn, giraffe, turtle");
+console.log(favoriteAnimals);
+console.log("The array has a length of: " + favoriteAnimals.length);
+favoriteAnimals.splice(3, 1);
+console.log(favoriteAnimals);
+console.log("The item you are looking for is at index: " + favoriteAnimals.indexOf("meerkat"));
```
yash-kapila, line 19: This is correct but you will learn a little later that using splice actually changes the favoriteAnimals array itself. This is called mutating the data structure and often it is suggested that we don't mutate the data structure i.e try to push the meerkat in between the two elements but return a new array itself thus having two arrays at the end, original and the modified array.

Maybe when you are introduced to this concept, you can revisit it and think of a way by which you can still add meerkat but don't modify the existing array. 🙂

orhan-homework-week3.js
```diff
@@ -0,0 +1,167 @@
...
+
+
+//3. Part More JavaScript!
+//1.
+function sum(a, b, c) {
+  return a + b + c;
+}
```
yash-kapila, line 26: A pure function 👍

orhan-homework-week3.js
```diff
@@ -0,0 +1,167 @@
...
+
+//2.
+function colorCar(color) {
+  console.log(`a ${color} car`);
```
yash-kapila, line 30: 👍 for using template strings here

orhan-homework-week3.js
```diff
@@ -0,0 +1,167 @@
...
+}
+colorCar("red");
+
+//3.
+function makeUser(newObj) {
+  console.log(newObj);
```
yash-kapila, line 36: Hi Orhan, I think the assignment was looking for you to loop over the object properties and print them instead of printing the whole object here. Do you remember from our class `for-in` loops or `Object.keys()`?

orhan-homework-week3.js
```diff
@@ -0,0 +1,167 @@
...
+};
+makeUser(obj);
+
+//4.
+let vehicleCode = {
+  1: "car",
+  2: "motorbike",
+};
+function vehicleType(color, code) {
+  console.log("a " + color + " " + vehicleCode[code]);
```
yash-kapila: Why not use template strings here like you did above?

orhan-homework-week3.js
```diff
@@ -0,0 +1,167 @@
...
+}
+vehicleType("red", 1);
+
+//5.
+const withoutIf = 3 === 3 ? console.log("yes") : console.log("no");
```
yash-kapila, line 56: 👍

orhan-homework-week3.js
```diff
@@ -0,0 +1,167 @@
...
+
+//6.
+function vehicle(color, code, age) {
+  if (age > 0) {
+    age = "used";
+  } else {
+    age = "new";
+  }
+  console.log("a " + color + " " + age + " " + vehicleCode[code]);
```
yash-kapila: Template strings here again?

orhan-homework-week3.js
```diff
@@ -0,0 +1,167 @@
...
+
+//7.
+let vehicles = ["motorbike", "caravan", "bike"];
+console.log(vehicles);
+//8.
+console.log("Third element of vehicles array is: " + vehicles[2]);
+
+//9.
+function vehicle(color, code, age) {
+  vehicleCode["3"] = "bike";
```
yash-kapila: Hi Orhan, I think you misunderstood the requirement here. You are supposed to make use of the list you created for Q.7 in this function instead of using the `vehicleCode` object 🙂

orhan-homework-week3.js
```diff
@@ -0,0 +1,167 @@
...
+    vehicles2[i] = (vehicles[i] + "s");
+  }
+  for (i = 0; i < vehicles2.length - 2; i++) {
+    vehicles2[i] += ", ";
+  }
+  for (i = vehicles2.length - 2; i < vehicles2.length - 1; i++) {
+    vehicles2[i] += " and ";
+  }
+  vehicles2[vehicles2.length - 1] += ".";
+  console.log(myStr + vehicles2.join(""));
```
yash-kapila, line 102: A few things here:
1. FYI - `vehicles.unshift("car")'` is an impure function i.e it mutates the array on which it is called. Even though it doesn't have a significant impact in our situation, it is not recommended to use such functions. But this is something you will learn later. 🙂 
2. The first `for` loop has got the correct declaration `let i = 0;` but the other loops don't. Can you explain the reason please?
3. Can we perhaps use only a single loop but have if-else condition inside that loop to append commas, and and the full stop rather than looping thrice? Just an alternative which you can think about

orhan-homework-week3.js
```diff
@@ -0,0 +1,167 @@
...
+}
+myFunc();
+
+//11. 
+vehicles.push("new vehicle");
```
yash-kapila, line 107: Ah. This will then cause issues with Q10, wouldn't it? And that's what the question asks to make sure that it doesn't happen 😉

orhan-homework-week3.js
```diff
@@ -0,0 +1,167 @@
...
+
+//13.
+const teachers = {
+  name: "Philipp, Rob, Unmesh, Yash",
+};
+console.log(teachers.name);
+
+//14.
+teachers.language = "HTML, CSS, JS";
+console.log(teachers.language);
```
yash-kapila, line 122: 👍

yash-kapila (commented):
Hi @OrhanGoc - apologies for leaving this a little late but I have reviewed your PR. However, before approving the same, I would like to ask you to look into the comments I have put in and provide an update please. Have a good day ahead and see you tomorrow.

OrhanGoc:
Hi @yash-kapila,
Thank you for your review, it is really helpful to understand the nuances that we should take into account while coding.
I'm going to update but is it enough to push a new commit to update or do we need a new PR?

orhan-homework-week3.js
OrhanGoc: Hi @yash-kapila,
I changed my assignment as you recommend

PR #92, opened 2019-03-13
Some loops, objects and functions of JavaScript1

ghifo:


homework-week3.js
```diff
@@ -0,0 +1,196 @@
+// 1. Strings!
+
+let myString = "hello,this,is,a,difficult,to,read,sentence";
+console.log(myString);   //hello,this,is,a,difficult,to,read,sentence
+console.log(myString.length);   //42
+
+//Find a way to remove the commas from the string and replace them with spaces.
+for (let i=0; i<myString.length; i++){
```
paulbremer, line 8: While the output is correct, there is no need to loop over the length of the string. If you look in the console log the output is 42 times the same string. Removing the loop will result in the same output.

homework-week3.js
```diff
@@ -0,0 +1,196 @@
...
+
+// 3 arguments function that returns the sum
+function sum (a,b,c) {
+    return a+b+c;
+}
+sum (1,2,3);   // 6
+
+//---------colorCar function -----------
+function colorCar (color) {
+    return "a "+ color + " car";   //the input should be a string .... I tried a lot but couldn't use red as a color not as a string????
```
paulbremer, line 55: The function works as it should, it returns 'a red car', do you still have a question about this task?

homework-week3.js
```diff
@@ -0,0 +1,196 @@
...
+
+/*
+if (3 === 3) {
+  console.log("yes");
+} else {
+  console.log("no");
+}
+*/
+const message = (3 === 3) ? "yes" : "no";
+console.log(message);  //      yes
```
paulbremer, line 88: Very good, this could be a one-liner as well, `console.log(3 === 3 ? 'yes' : 'no')`

homework-week3.js
```diff
@@ -0,0 +1,196 @@
...
+
+//--------vehicle function (color, code, age)---------
+function vehicle (color, code, age){
+    if(age > 1 && code === 1){
+        return "a " + color + " used" + " car";
```
paulbremer, line 93: The use of the " + " can become quite hard, take a look at the new backticks in which you can write ``a ${color} used car``
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

homework-week3.js
```diff
@@ -0,0 +1,196 @@
...
+
+vehicle("green", 3, 1)    //"a green new bike"
+
+// -------plurals/ commas + space---------
+//const listOfVehicles = ["motorbike", "car", "caravan", "bike"];     (mentioned above)
+function loopFunction(listOfVehicles){
+    for (let i=0; i < listOfVehicles.length; i++){
+       listOfVehicles[i] = listOfVehicles[i] + "s";
+    }
+    clearListOfVehicles = listOfVehicles.join(', ');   // I couldn't replace the (,) with (and), how?!!!!
```
paulbremer, line 132: It takes a bit more work, you need to count the items in the loop and check if it's the last item, if it is the last item put `and` in front of it.

homework-week3.js
```diff
@@ -0,0 +1,196 @@
...
+console.log("the order that you assign (o3 = o2 or o2 = o3) matters because (o2 = o3) will gives syntax error.");
+
+/*What does the following code return? (And why?)
+
+let bar = 42;
+typeof typeof bar;
+‘Coerce' means to try to change - so coercing var x = '6' to number means trying to change the type to number temporarily.*/
+let bar = 42;
+typeof typeof bar;     //"string"
+console.log("It will return 'string' because it returns the typeof bar as a value for the second typeof expression");
```
paulbremer, line 196: Yes! 👍

paulbremer (approved):
Hi Ghufran, good job! Do let me know if you have any more questions :)

homework-week3.js
ghifo: I know it is toooooo late, but I could not sole the 'and ' problem

paulbremer: You are inside the for loop so you have access to the `i` which holds the index, for every loop you know the `index`, next you want to know the length of the array, which you also know `listOfVehicles.length`. 

The next thing to do is check if the `index` is the same as the length of the array minus one. If that is the case join with 'and' otherwise join with ','.

ghifo:
I’ll try that, thanks a lot 😊

PR #93, opened 2019-03-14
Added homework week 3

Nadoshka:


Week2/nadin.js
```diff
@@ -0,0 +1,181 @@
+'use strict'
```
paulbremer: The homework is in the week2 folder but the it is week3 already!

Week2/nadin.js
```diff
@@ -0,0 +1,181 @@
...
+
+//strings
+
+let myString = "hello,this,is,a,difficult,to,read,sentence";
+console.log(myString);
+console.log(myString.length); 
+
+const space = " ";
+myString = 'hello' + space + 'this' + space + 'is' + space + 'a' + space + 'difficult' + space + 'to' + space + 'read' + space + 'sentence';
+console.log(myString);
```
paulbremer: This works but it is not really effective, take a look again at `replace` and  `split` . These functions work to split up a string. For example `myString.replace(/,/g, ` `)`, where you replace the `,` character with a space.

Week2/nadin.js
```diff
@@ -0,0 +1,181 @@
...
+
+function sumThreeNumber (num1, num2, num3) {
+  return num1+num2+num3;
+}
+const total = sumThreeNumber(1,2,3);
+console.log(total);
+
+//2
+function colorCar (color) {
+return 'a' + ' ' + color + ' ' + 'car';
```
paulbremer: You can add the spaces inside the strings, `a ' + color + ' car`.

Week2/nadin.js
```diff
@@ -0,0 +1,181 @@
...
+//3
+
+const studentObj = {
+    name :"atik",
+    course :"programing",
+    number :'20', 
+}
+
+function printProperties(obj) {
+    for (const key in obj) {
```
paulbremer: Good use of the for loop!

Week2/nadin.js
```diff
@@ -0,0 +1,181 @@
...
+        console.log(key, ":" , obj[key]);
+      }
+}
+printProperties(studentObj);
+
+//4
+function vehicleType(color, code) {
+   switch (code) {
+       case 1: 
+       console.log('a' + ' ' + color + ' ' +  'car');
```
paulbremer: Look at the comment on task 2.

Week2/nadin.js
```diff
@@ -0,0 +1,181 @@
...
+
+function vehicle(color, code, age) {
+    if (age === 5 && code === 1){
+        console.log('a' + ' ' + color + ' ' + 'used car');
+        return;
+    } else if (age === 5 && code === 2){
+        console.log('a' + ' ' + color + ' ' + 'used motorbike');
+        return;
+    }
+    switch (code) {
```
paulbremer: Because you have `return` in the if statement above this switch code is never reached, you can remove the code here.

Week2/nadin.js
```diff
@@ -0,0 +1,181 @@
...
+
+vehicleType("blue", 2);
+
+//5
+3 === 3 ? console.log ('yes') : console.log('no');
+
+//6
+
+function vehicle(color, code, age) {
+    if (age === 5 && code === 1){
```
paulbremer: What if the age is not 5, but the age is 1? Can you imagine a solution that takes care of different ages like 1, 3 or 5?

Week2/nadin.js
```diff
@@ -0,0 +1,181 @@
...
+        console.log('a' + ' ' + color + ' ' + 'new' + ' ' + vehicles[code-1]);
+    } else {
+        console.log('a' + ' ' + color + ' ' + 'used' + ' ' + vehicles[code-1]);
+    } 
+ }
+ //10
+const vehicles = ["motorbike", "caravan", "bike"];
+let advertisement = "Amazing Joe's Garage, we service cars, "
+for (let i = 0; i < vehicles.length; i++) {
+     if(i === vehicles.length - 1) {
```
paulbremer: Very good!

Week2/nadin.js
```diff
@@ -0,0 +1,181 @@
...
+         advertisement += "and " + vehicles[i] + "s";
+     } else {
+        advertisement += vehicles[i] + "s, ";
+     }
+ }
+console.log(advertisement);
+
+//11
+const vehicles = ["motorbike", "caravan", "bike"];
```
paulbremer: The task is to add a vehicle without changing the code, so while this works there is a better way. Rewrite the task 10 code to take in an array of vehicles and if that works, only `vehicles.push("bus");` should be enough to make a new advertisement.

Week2/nadin.js
```diff
@@ -0,0 +1,181 @@
...
+    console.log('x == y', x == y, "they do not have the same reference");
+    console.log('x === y', x === y, "they do not have the same reference");
+    console.log("z == y", z == y, "they have the same reference" );
+    console.log('z == x', z == x, "they do not have the same reference");
+
+//16
+let o1 = { foo: "bar" };
+let o2 = { foo: "bar" };
+let o3 = o2;
+console.log({o1}, {o2}, {o3});
```
paulbremer: `o1` is already an object, there is no need to wrap `o1` in an object in the console.log(). So `console.log(o1, o2, o3)` works just fine.

Week2/nadin.js
```diff
@@ -0,0 +1,181 @@
...
+o3 = o2; 
+o2 = o3;
+console.log("It dose matter because o3 is undefined and 02 is already declared");
+
+//17
+let bar = 42;
+typeof typeof bar;
+
+console.log(typeof typeof bar);
+console.log("typeof typeof bar: is string because we asked about the type of the word 'bar' which is string not about the type of the value of the variable bar which is number");
```
paulbremer: 👍

paulbremer (approved):
Nice work! Some small comments, if you have any questions please let me know :)

Week2/nadin.js
Nadoshka: Thanks for the heads up! I will keep it in mind for the next time. :)

Week2/nadin.js
Nadoshka: I just want to use switch code to practice the q on another way, Thanks for the info! :)

Week2/nadin.js
Nadoshka: Yes,  It could be >= 5! 
Thanks! That will be handlier and good for me to practice it on this way.

Week2/nadin.js
Nadoshka: That is easier! Thanks 👍

Week2/nadin.js
Nadoshka: It's done! Thanks,I will keep it in mind! :)

Week2/nadin.js
Nadoshka: Thanks! That help me well to do it on easy way and also it is shorter. :) 👍

Week2/nadin.js
Nadoshka: I want to see the info of the object on the console of browser.
Thanks for the feedback! :)

Week2/nadin.js
Nadoshka: Thanks for your feedback! It is done! 👍 :)

Week2/nadin.js
Nadoshka: I just want to use switch code to practice the q on another way, Thanks for the info! :)

PR #94, opened 2019-03-15
Semih homework 2

sigma21:
Although I haven't made any changes to the previous week's file, GitHub is still adding it as well. (displays "2 files changed" message)

### DIFF
--- a/semih2.js
+++ b/semih2.js
@@ -1,0 +1,190 @@
+// #1
+let myString = "hello,this,is,a,difficult,to,read,sentence";
+console.log(myString);
+console.log(myString.length);
+
+const newString = myString.split(",").join(" ");
+console.log(newString);
+
+
+// #2
+let favoriteAnimals = ["blowfish", "capricorn", "giraffe"];
+favoriteAnimals.push("turtle");
+console.log(favoriteAnimals);
+
+favoriteAnimals.unshift("meerkat");
+console.log(favoriteAnimals);
+console.log("The array has a length of: ", favoriteAnimals.length);
+
+favoriteAnimals.splice(3,1);
+console.log(favoriteAnimals);
+console.log("The item you are looking for is at index: ", favoriteAnimals.indexOf("meerkat"));
+
+favoriteAnimals.splice(favoriteAnimals.indexOf("meerkat"),1);
+console.log(favoriteAnimals);
+
+
+// #3
+//////////////1//////////////////////
+function addUp (x,y,z) {
+    const totalSum = (x+y+z);
+    console.log(totalSum);
+}
+addUp(1,10,100);
+/////////////////////////////////// 
+
+
+///////////////2///////////////////
+function colorCar(color) {
+    console.log("A", color, "car.");
+}
+colorCar("black");
+///////////////////////////////////
+
+
+///////////////3//////////////////
+const myObj = {name:"john doe", age:"24", city:"amsterdam"}
+
+function displayValue(object) {
+    console.log(object.name, object.age, object.city);
+}
+displayValue(myObj);
+////////////////////////////////
+
+
+////////////4////////////////////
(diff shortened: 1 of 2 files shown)


semih2.js
```diff
@@ -0,0 +1,190 @@
...
+const newString = myString.split(",").join(" ");
+console.log(newString);
+
+
+// #2
+let favoriteAnimals = ["blowfish", "capricorn", "giraffe"];
+favoriteAnimals.push("turtle");
+console.log(favoriteAnimals);
+
+favoriteAnimals.unshift("meerkat");
```
paulbremer, line 15: unshift adds the value to the beginning of the array, but it needs to go after blowfish. You can use `splice` for this.

semih2.js
```diff
@@ -0,0 +1,190 @@
...
+    const totalSum = (x+y+z);
+    console.log(totalSum);
+}
+addUp(1,10,100);
+/////////////////////////////////// 
+
+
+///////////////2///////////////////
+function colorCar(color) {
+    console.log("A", color, "car.");
```
paulbremer, line 39: This works but can be done in some other ways, `'A ' + color + ' car.'` or using backticks `a ${color} car`

semih2.js
```diff
@@ -0,0 +1,190 @@
...
+    }
+}
+vehicle("red",1,0);
+vehicle("white",2,1);
+//////////////////////////////////
+
+
+//////////////7-8//////////////////
+const vehicles = ["motorbike", "caravan", "bike", "automobile", "truck"];
+const thirdElement = vehicles[2];
```
paulbremer, line 96: You are not using this const thirdElement.

semih2.js
```diff
@@ -0,0 +1,190 @@
...
+vehicle("green",3,1);
+/////////////////////////////////
+
+
+//////////////10////////////////
+const vehicleList = ["motorbike", "caravan", "bike", "automobile", "truck"];
+let advertisement = "Amazing Joe's Garage, we service " 
+
+for (let i=0; i<vehicleList.length-1; i++) {
+    advertisement += vehicleList[i]+", ";
```
paulbremer, line 119: You can add a `s` after the vehicle here, so `bikes, caravans`.

semih2.js
```diff
@@ -0,0 +1,190 @@
...
+}
+
+advertisement += "and " + vehicleList[vehicleList.length-1] + ".";
+console.log(advertisement);
+//////////////////////////////
+
+
+/////////////11///////////////
+// const vehicleList = ["motorbike", "caravan", "bike", "automobile", "truck", "one more vehicle"];
```
paulbremer, line 128: No need to comment this out, this code works fine :)

paulbremer (approved):
Hi Semih! Good job on the homework, some small comments. Let me know if you have any questions.

semih2.js
sigma21: Right, I misread it somehow and thought I should add to the beginning. I should have used `favoriteAnimals.splice(1, 0, 'meerkat');`

semih2.js
sigma21: Should have console logged `thirdElement` instead of `vehicles[2]` . :)

semih2.js
sigma21: Right, makes sense. Thanks.

semih2.js
sigma21: I commented that part out to check it in browser console. I defined `vehicleList` one more time somewhere in the .js file so it would throw me an error for defining the same variable twice.
You're right, it works fine :)

semih2.js
sigma21: Thanks for the alternatives! 👍  I guess I should get used to using `$` sign more often :)

sigma21:
Thanks a lot @paulbremer for the review and feedback!

PR #95, opened 2019-03-15
Homework JavaScript1 Week 3

RamazanInce:


Week2/string.js
```diff
@@ -0,0 +1,8 @@
+console.log('String and Array challenges');
+console.log(`1. Strings!`); // how to remove commas in a string
+let myString = 'hello,this,is,a,difficult,to,read,sentence';
+console.log(myString);
+console.log(myString.length);
+myString = myString.split(',');
+myString = myString.join(' ');
```
paulbremer, line 7: You can chain functions like this; `myString.split(',').join(' ')`

Week2/arrays.js
```diff
@@ -0,0 +1,12 @@
+console.log(` 
+2. Arrays!`); // how to add or remove input in an array.
+let favoriteAnimals = ['blowfish', 'capricorn', 'giraffe'];
+favoriteAnimals.push('turtle');
+console.log(favoriteAnimals);
+favoriteAnimals.shift();
+favoriteAnimals.unshift('blowfish', 'meerkat');
```
paulbremer, line 7: This works, but can you find a way to use the `splice` function?

Week2/task1.js
```diff
@@ -0,0 +1,7 @@
+console.log(`More JavaScript`);
+console.log(`
+Task 1`); // a function which adds three given arguments.
+function sum(a, b, c) {
+  return (sum = a + b + c);
```
paulbremer, line 5: You don't need to call the function in here, `return a + b + c` will work just fine.

Week2/task10.js
```diff
@@ -0,0 +1,28 @@
...
+let advert = "Amazing Joe's Garage, we service ";
+for (let index = 0; index < listOfVehicles.length - 2; index++)
+  advert = advert + listOfVehicles[index] + 's, ';
+advert =
+  advert +
+  listOfVehicles[listOfVehicles.length - 2] +
+  's' +
+  ' and ' +
+  listOfVehicles[listOfVehicles.length - 1] +
+  's.';
```
paulbremer, line 27: Nice!

Week2/task15.js
```diff
@@ -0,0 +1,25 @@
...
+console.log(`
+Task 15 `); // testing the equality of the same arrays
+let x = [1, 2, 3];
+let y = [1, 2, 3];
+let z = y;
+console.log('I think x == y is true');
+console.log('I think x === y is true');
+console.log('I think z == y is true');
+console.log('I think z == x is true');
+
```
paulbremer, line 11: Only one of these is actually true, have you checked all of them and do you understand why only 1 is true?

Week2/task16.js
```diff
@@ -0,0 +1,19 @@
...
+o2 = { animal: 'cat' };
+
+console.log(o3);
+console.log('no');
+
+o1 = { animal: 'dog' };
+console.log(o3);
+console.log('no');
+
+console.log('we can not write o2 = o3, since we already define o2 and we did not define o3 yet');
```
paulbremer, line 18: You defined `o3` on line 7, so we can write `o2 = o3`, but if we change the value of `o2` will it change `o3` as well?

Week2/task4.js
```diff
@@ -0,0 +1,16 @@
...
+    code = 'car';
+  } else if (code === 2) {
+    code = 'motorbike';
+  } else {
+    code = 'but vehicle type is unknown';
+  }
+  console.log('a ' + color + ' ' + code);
+}
+
+vehicleType('blue', 3);
```
paulbremer, line 16: Did you try this line? It returns `a blue but vehicle type is unknown`. The code is not wrong but the result is a bit weird.

Week2/task6.js
```diff
@@ -0,0 +1,22 @@
...
+    age = 'used';
+  } else if (age === 1) {
+    age = 'brand new';
+  } else {
+    age = 'wrong input';
+  }
+  console.log('a ' + color + ' ' + age + ' ' + code);
+}
+
+vehicle('blue', 2, 0);
```
paulbremer, line 22: If the age is 0, shouldn't it be a brand new car?

Week2/task9.js
```diff
@@ -0,0 +1,28 @@
...
+    age = 'used';
+  } else if (age === 1) {
+    age = 'brand new';
+  } else {
+    age = 'wrong input';
+  }
+  console.log('a ' + color + ' ' + age + ' ' + listOfVehicles[code]);
+}
+
+vehicleNew('blue', 2, 1);
```
paulbremer, line 28: 👍

paulbremer (approved):
Very nice Ramazan! Take a look at the comments and let me know if you have any questions.

Week2/string.js
RamazanInce: That is really nice.
Thanks

Week2/arrays.js
RamazanInce: Yes
instead of Line 6 we can use
favoriteAnimals.splice(0, 1);

Week2/task1.js
RamazanInce: Yes, I did it now.
Thanks.

Week2/task15.js
RamazanInce: I try to write a function which shows if two arrays have the same element then the function will return to true. That is what I understood from the question. I observe that with writing the equality to the console.

Week2/task16.js
RamazanInce: On my understanding 
I should write line 7 instead of o3 = o2 twrite o2 = o3, that is why I gave an explanation like that.
No.

Week2/task4.js
RamazanInce: Since we have only code 1 and 2 I try to put extra explanation, for the cases where the code will be different than 1 and 2. For 1 and 2 it works.

Week2/task6.js
RamazanInce: Yes, I think so. What do you think or just new?

Week2/task9.js
RamazanInce: Thanks

RamazanInce:
@paulbremer 
Thank you very much for your feedback's. Those points assist me a lot and add value to my knowledge.
I will my self luckily since I have a teachers like you. 👍

PR #98, opened 2019-03-18
Omar Qaterge homework week3

omarqaterge:


omar_qaterge.js
```diff
@@ -0,0 +1,202 @@
+//1
```
paulbremer, line 1: This is the homework from week2, try to include only the homework for the current week in the pull request. You can make a new branch from master and work from there.

week3.js
```diff
@@ -0,0 +1,223 @@
...
+
+// 2 Arrays:
+let favoriteAnimals = ["blowfish", "capricorn", "giraffe"];
+favoriteAnimals.push("turtle");  // adding Mauro's favorite animal at the end of the previous array
+console.log(favoriteAnimals);
+favoriteAnimals.splice(1, 0, "meerkat");
+console.log("using splice() a method that changes the contents of an array by removing or replacing existing elements and/or adding new elements)\n \
+in .splice(1, 0, 'meerkat') the 1 defines the index that i want to work with and 0 defines the number of elements that i want to replace.\n \
+so the value of the array will be:\n \
+[\"blowfish\", \"meerkat\", \"capricorn\", \"giraffe\", \"turtle\"]");
```
paulbremer, line 21: Nice explanation!

week3.js
```diff
@@ -0,0 +1,223 @@
...
+console.log(favoriteAnimals);
+console.log("The array has a length of: " + favoriteAnimals.length);
+favoriteAnimals.splice(3,1);   // delete giraffe from the previous array
+console.log(favoriteAnimals);
+console.log("The item you are looking for is at index: " + favoriteAnimals.findIndex(animal => animal === "meerkat"));
```
paulbremer, line 26: This works but a better function is to use the `indexOf` if you want to search for a simple value, `indexOf('meerkat')`. You can read more about the differences of findIndex and indexOf here https://stackoverflow.com/questions/41443029/difference-between-indexof-and-findindex-function-of-array

week3.js
```diff
@@ -0,0 +1,223 @@
...
+
+
+
+
+// More JavaScript
+
+//1
+function sum(x, y, z) {
+  const result = x + y + z;
+  return result;
```
paulbremer, line 36: You can even return the statement and don't need to make a const; `return x + y + z`

week3.js
```diff
@@ -0,0 +1,223 @@
...
+     	id2: "helicopter",
+     }
+}
+const firstKey = Object.keys(vehicle)[0];
+const secondKey = Object.keys(vehicle)[1];
+const thirdKey = Object.keys(vehicle)[2];
+//----------------------------------------------------
+console.log(vehicle["aircraft"]["id2"]);
+//logs helicopter
+console.log(vehicle[thirdKey]["id2"]);
```
paulbremer, line 114: Nice!

week3.js
```diff
@@ -0,0 +1,223 @@
...
+  }
+  else if(code === 2){
+  	i = vehicles[1];
+  }
+  else if(code === 3){
+  	i = vehicles[2]
+  } 
+  else if(code === 4){
+  	i = vehicles[3]
+  }
```
paulbremer, line 147: This is fine but you can already see that it becomes a bit cumbersome to write almost the same code over and over again. With larger if statements you can look if a loop is better suited for the job.

week3.js
```diff
@@ -0,0 +1,223 @@
...
+console.log(vehicles);
+let vehicleToService="";
+for(let counter = 0; counter < vehicles.length; counter ++){
+  if(counter === vehicles.length - 1){
+  	vehicleToService += vehicles[counter] + "s.";
+  } 
+  else {vehicleToService += vehicles[counter] + "s, ";}
+ }
+
+console.log(" \"Amazing Joe's Garage, we service" , vehicleToService,"\"");
```
paulbremer, line 188: There is no need for the slashes here `console.log("Amazing Joe's Garage, we service" , vehicleToService);`, this works fine. You could even use backticks which makes it even easier to work with variables and strings, see https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Template_literals

week3.js
```diff
@@ -0,0 +1,223 @@
...
+let o3 = o2;
+
+console.log(o3);
+o2.foo = "something";
+console.log(o3);
+o1.foo = "and something else";
+console.log(o3);
+
+//Does the order that you assign (o3 = o2 or o2 = o3) matter?
+//yes, because the right side is being assigned to the left side, otherwise it will show that the variable is not defined.
```
paulbremer, line 214: Yes that is correct. You changed `foo` on the `o2` object, try this exercise again but this time change the whole object so `o2 = { foo: "something" }`. See if that changes anything.

week3.js
```diff
@@ -0,0 +1,223 @@
...
+
+//12
+const object = {};
+
+
+
+//13
+const person = {htmlCss:"Philipp Beau", cli:"Unmesh Joshi", JavaScript:"Yash Kapila"};
+
+
```
paulbremer, line 200: Did you forgot task 15?

week3.js
```diff
@@ -0,0 +1,223 @@
...
+//yes, because the right side is being assigned to the left side, otherwise it will show that the variable is not defined.
+
+
+
+//What does the following code return? (And why?)
+let bar = 42;
+typeof typeof bar;
+
+//typeof bar; returns "number" because a number is assigned to bar as a value
+//typeof typeof bar; returns "string" because it is returning the type of "number" which is a string
```
paulbremer, line 223: 👍

paulbremer (changes requested):
Good job Omar! Some small comments and it looks like you forgot task 15? 

Task 15: Write some code to test two arrays for equality using `==` and `===`.  

Take a look at the code and check if arrays are the same using `==` and `===`. Try console.logging the output; `console.log(x == y); // false`

PR #102, opened 2019-04-28
homework of javascript week2

aalhommada:


Week2/num4.js
```diff
@@ -0,0 +1,12 @@
+// 4
+let y = "Abdullah is doing homework";
+// 4.1
+console.log("the value of variabl y will be : Abdullah is doing the homework")
```
marcianoviereck: Here you wrote:
"Abdullah is doing **the** homework"

But I think the value is:
"Abdullah is doing homework".

It's a small thing!

Week2/num4.js
```diff
@@ -0,0 +1,12 @@
...
+// 4.2
+console.log(y)
+// 4.3 
+y = "I'm doing it perfectly";
+// 4.4
+console.log("the value of variable y will be : I'm doning that perfectly")
```
marcianoviereck: Here it's the same:

"I'm **doning** **that** perfectly"

Which is not the same as the value:)

Week2/num4.js
marcianoviereck: And you forgot to add a ';' at the end of this line (which you seem to have forgotten at all `console.log` statements

Week2/num5.js
```diff
@@ -0,0 +1,18 @@
+// 5
+console.log(Math.round(7.25));
+// 5.1
+let z = 7.25 ;
```
marcianoviereck: You added a 'space' here, which is not necessary.

So you can just write:

`let z = 7.25;`

Week2/num5.js
```diff
@@ -0,0 +1,18 @@
...
+// 5.2
+console.log(z)
```
marcianoviereck: You forgot the ';'

Week2/num5.js
```diff
@@ -0,0 +1,18 @@
...
+// 5.3
+let a = Math.round(z)
+// 5.4
+console.log(a)
+// 5.5 
+let highValue ;
+if (a > z) {
+    highValue = a;
+} else { highValue = z}
```
marcianoviereck: I would suggest putting the else on a new line. Like this:

```
if (a > z) {
    highValue = a;
} else { 
    highValue = z;
}
```

marcianoviereck: Besides that it's good. Just a small tip, you can also use [Math.max](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max)! But this solution is fine!

Week2/num6.js
```diff
@@ -0,0 +1,15 @@
+// 6 Arrays
+// 6.1
+let Friends = [];
```
marcianoviereck: you don't have to use uppercase `F` here. If you start the variable name with a capital letter, it might suggest that it is a `class`.

So `let friends = [];` is better (see [naming conventions](https://github.com/HackYourFuture/fundamentals/blob/master/fundamentals/naming_conventions.md))

Week2/num7.js
```diff
@@ -0,0 +1,9 @@
+// 7 . More strings
+let myString = "this is a test";
+// 7.1
+console.log(myString);
+// 7.2
+let myStringLength = myString.length ;
```
marcianoviereck, line 6: nice!

Week2/num8.js
```diff
@@ -0,0 +1,44 @@
+// 8
+
+// 8.1
+let fooString = "100";
+let fooNumber = 100 ;
+let fooArray = [1, 2, 3];
+let fooX ;
```
marcianoviereck: I think the assignment wants you to use  a `boolean` here, since `undefined` is not a variable type, but `boolean` is. So could you change that?

Week2/num8.js
```diff
@@ -0,0 +1,44 @@
...
+// 8.2
+console.log("the value of my variable String is: " + fooString)
```
marcianoviereck: I would recommend always adding ';' at the end of a line of code (or function)! Because it might lead to some silly mistakes later on

Week2/num8.js
```diff
@@ -0,0 +1,44 @@
...
+
+// 8.4
+console.log(typeof fooString);
+console.log(typeof fooNumber);
+console.log(typeof fooArray);
+console.log(typeof fooX);
+
+// 8.5
+// 8.6
+if (fooString===fooNumber){
```
marcianoviereck: It's better to have a 'space' in between the if statement:

`if (fooString === fooNumber) {`

Week2/num8.js
```diff
@@ -0,0 +1,44 @@
...
+} else {
+  console.log("NOT THE SAME TYPE")
+  console.log(" fooString type is : "+typeof fooString +" , while fooArray type is : " + typeof fooArray)
+}
+if (fooString===fooX){
+  console.log("SAME TYPE");
+} else {
+  console.log("NOT THE SAME TYPE")
+  console.log(" fooString type is : "+typeof fooString +" , while fooNumber type is : " + typeof fooX)
+}
```
marcianoviereck, line 66: I think you forgot to compare `fooNumber` with `fooArray`
And `fooNumber` with `fooX` and `fooX` with the `fooArray`

Week2/num9.js
```diff
@@ -0,0 +1,14 @@
+// 9
+let x = 7;
+console.log("the new value of x after applying x = x % 3   is : 1 "+ " , 7 will divided by 3 and what is remaining is the new value")
```
marcianoviereck: You don't need to add the `+` here, you can just write it in one string:)

Week2/num10.js
```diff
@@ -0,0 +1,25 @@
...
+let y = 10/0;
+if (x===y){
+  console.log("comparing is possible because " + x + "=" + y);
+} 
+if (x > y){
+  console.log("6/0 is greater than 10/0");
+}
+if (x < y) {
+  console.log("10/0 is greater than 6/0")
+}
```
marcianoviereck, line 25: Nice!

Tip: If you want to check if something is equal or smaller than something else, you can use
```
if (x <= y) {
  console.log("10/0 is greater than 6/0")
}
```
Or if something is equal or larger than something else:
```
if (x >= y) {
  console.log("10/0 is greater than 6/0")
}
```
Some more are [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators)

marcianoviereck (changes requested):

Week2/num1.js
```diff
@@ -0,0 +1,13 @@
+
+let greeting = "Hellow World" // greeting in many languages
```
marcianoviereck, line 2: Overall the code looks good! There are just some small things here and there!

Keep it up!

Week2/num8.js
marcianoviereck: Ah and you forgot to use `typeof` to compare the types. Now you are comparing the `values` of the `variables`

marcianoviereck (changes requested):

Week2/num8.js
aalhommada: I just merge this step with next one ( when I did the comparison )

marcianoviereck (approved):
Nice! Good fixes and the homework is approved now!

PR #103, opened 2019-04-28
abdullah github-username:heartwhite

heartwhite:
Javascript week1homework

### DIFF
--- a/Week1/arrays.js
+++ b/Week1/arrays.js
@@ -4,5 +4,5 @@ console.log ('my array will contain some animals');
 console.log(arraysInside);
 let myFavoriteAnimals = ['eagle ', 'dog ', 'horse '];
 console.log(myFavoriteAnimals);
-arraysInside.push('baby pig');
-console.log(arraysInside);
+myFavoriteAnimals.push('baby pig');
+console.log(myFavoriteAnimals);
--- a/Week1/arrays.js
+++ b/Week1/arrays.js
@@ -1,0 +1,8 @@
+//append arrays
+let arraysInside = [];
+console.log ('my array will contain: eagle, horse and dog');
+console.log(arraysInside);
+let myFavoriteAnimals = ['eagle ', 'dog ', 'horse '];
+console.log(myFavoriteAnimals);
+arraysInside = ['baby pig ' , myFavoriteAnimals];
+console.log(arraysInside);
--- a/Week1/arrays.js
+++ b/Week1/arrays.js
@@ -1,8 +1,8 @@
 //append arrays
 let arraysInside = [];
-console.log ('my array will contain: eagle, horse and dog');
+console.log ('my array will contain some animals');
 console.log(arraysInside);
 let myFavoriteAnimals = ['eagle ', 'dog ', 'horse '];
 console.log(myFavoriteAnimals);
-arraysInside = ['baby pig ' , myFavoriteAnimals];
+arraysInside.push('baby pig');
 console.log(arraysInside);
--- a/Week1/checkType.js
+++ b/Week1/checkType.js
@@ -1,0 +1,48 @@
+//using the 'typeof' and compairing types with if statement
+const myName = 'abdullah';
+let myAge = 33;
+let myHobbies = ['swimming', 'coding', 'taking photographs'];
+let iAmFunny = true;
+let amIFunny = 'yes';
+console.log(myName);
+console.log(myAge);
+console.log(myHobbies);
+console.log(iAmFunny);
+console.log(amIFunny);
+console.log('types of my variables are: string, number, array, boolean, string \(again\)');
+console.log(typeof myName);
+console.log(typeof myAge);
+console.log(typeof myHobbies);
+console.log(typeof iAmFunny);
+console.log(typeof amIFunny);
+
+
+if(typeof myName === typeof myAge){
+    console.log('SAME TYPE')
+} else{
+    console.log('NOT SAME TYPE')
+}
+
+if(typeof myAge === typeof myHobbies){
+    console.log('SAME TYPE')
+} else{
+    console.log('NOT SAME TYPE')
+}
+
+if(typeof myHobbies === typeof iAmFunny){
+    console.log('SAME TYPE')
+} else{
+    console.log('NOT SAME TYPE')
+}
+
+if(typeof iAmFunny === typeof amIFunny){
+    console.log('SAME TYPE')
+} else{
+    console.log('NOT SAME TYPE')
+}
+
+if(typeof amIFunny === typeof myName){
+    console.log('SAME TYPE')
+} else{
+    console.log('NOT SAME TYPE')
+}
--- a/Week1/checkType.js
+++ b/Week1/checkType.js
@@ -23,7 +23,43 @@ if(typeof myName === typeof myAge){
     console.log('NOT SAME TYPE')
 }
 
+if(typeof myName === typeof myHobbies){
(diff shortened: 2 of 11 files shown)


Week1/helloInDifLang.js
```diff
@@ -0,0 +1 @@
+console.log ('hello world  ' + '\nздравствуйте мира  ' + '\nmerhaba dunya');
```
marcianoviereck: Hey this looks good,
but I think that the assignment was to add a separate `console.log` for each language, so can you change that:)?

Week1/varXInteger.js
```diff
@@ -0,0 +1,7 @@
+//giving an integer value to variable x
+let x;
+console.log("the value of my variable x will be: my favorite number");
```
marcianoviereck: I think that you forgot to change this `my favorite number`, because that i not the value that `x` has.

Week1/varXInteger.js
```diff
@@ -0,0 +1,7 @@
...
+console.log(x);
+x = 775;
+console.log("the value of x will be my favorite number: 775");
```
marcianoviereck: You can remove `my favorite number` here, but the rest is good!

Week1/numberRounding.js
```diff
@@ -0,0 +1,10 @@
+//round and compair numbers
+let z = 7.25;
+console.log(z);
+let a = Math.round(z);
+console.log(a);
+let numbers = [a, z];
+numbers.sort(function(a, b) { return b - a; });
```
marcianoviereck: haha wow, cool solution!

That's really nice, there is an easier solution though (but not as cool): [Math.max](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max)

Week1/arrays.js
```diff
@@ -0,0 +1,8 @@
+//append arrays
+let arraysInside = [];
+console.log ('my array will contain: eagle, horse and dog');
```
marcianoviereck: Here the array does not contain those values **yet**, but it will contain them after line 5!

Week1/arrays.js
```diff
@@ -0,0 +1,8 @@
...
+console.log(arraysInside);
+let myFavoriteAnimals = ['eagle ', 'dog ', 'horse '];
+console.log(myFavoriteAnimals);
+arraysInside = ['baby pig ' , myFavoriteAnimals];
```
marcianoviereck: Ah, this will result in an array inside an array, but it won't add 'baby pig' to the current array.

So this is not what the assignment was about, but it is very nice that you know how to do this!

Week1/checkType.js
```diff
@@ -0,0 +1,48 @@
...
+    console.log('SAME TYPE')
+} else{
+    console.log('NOT SAME TYPE')
+}
+
+if(typeof amIFunny === typeof myName){
+    console.log('SAME TYPE')
+} else{
+    console.log('NOT SAME TYPE')
+}
```
marcianoviereck: This is nice! But I think you forgot to compare some  variables (such as `myAge` to `myName`)

Week1/compairingInfinities.js
```diff
@@ -0,0 +1,13 @@
+// store diffrent types of data in one array, and compare the infinitives
+let allTypesInsideArray = ['abdullah ', 33 , true]
```
marcianoviereck: don't forget the ';' here

Week1/compairingInfinities.js
```diff
@@ -0,0 +1,13 @@
...
+
+
+console.log('2 examples for compairing infinities')
+if(6/0 === 10/0){
+    console.log(true);
+} else{
+    console.log(false);
+}
+
+console.log(6/0 == 10/0);
```
marcianoviereck: here you use `==` and in the other if statement you use `===` is this done on purpose:)?

And it's better to store the results of `6/0` and `10/0` in variables and then compare them, it makes your code a bit more readable!

marcianoviereck (changes requested):
The code looks good and you already use some more advanced things (like sorting)! Cool, I do have some comments though:)

heartwhite:
Hi Marciano. Thank you for feed back. 
I did homework directly how I understand it at first look. 
Maybe its about my low level of English . Sorry.
I am still not sure about Exercise 6. I added baby pig (I discovered it after your feedback) to array which I declared in first step of assignment. I can add to myFavoriteAnimals array in my understanding I need to add the existing array(first one)
Can you say it if its wrong now?


And rounding and compering assignment. I prefer your way for now. *(Math.max)
Because I did not understand totally sorting.

Week1/compairingInfinities.js
heartwhite: I wanted to see both option (== / ===)
I updated with two variables. 
thanks

Week1/compairingInfinities.js
heartwhite: done

Week1/checkType.js
heartwhite: I compared all. updated. thanks

Week1/arrays.js
heartwhite: I added baby pig with .push

Week1/arrays.js
heartwhite: updated

Week1/numberRounding.js
heartwhite: I prefer your way. Math.max. updated

Week1/varXInteger.js
heartwhite: done

Week1/varXInteger.js
heartwhite: done.

Week1/helloInDifLang.js
heartwhite: console.log for each languages. done

Week1/arrays.js
```diff
@@ -1,8 +1,8 @@
 //append arrays
 let arraysInside = [];
-console.log ('my array will contain: eagle, horse and dog');
+console.log ('my array will contain some animals');
 console.log(arraysInside);
 let myFavoriteAnimals = ['eagle ', 'dog ', 'horse '];
 console.log(myFavoriteAnimals);
-arraysInside = ['baby pig ' , myFavoriteAnimals];
+arraysInside.push('baby pig');
```
marcianoviereck: It is almost correct! If you change `arraysInside.push('baby pig');` to `myFavoriteAnimals.push('baby pig');` then it is correct:)

Because now you're pushing the 'baby pig' in the `arraysInside` array (which is empty). So then that array will **only** have 'baby pig' inside of it. 
But if you push it in the `myFavoriteAnimals` array, then that array has all the animals it already had ('eagle ', 'dog ', 'horse ') and it will also have 'my baby pig'.

marcianoviereck (changes requested):
The rest is good! If you fix this last thing then it is correct:)

So if you fix this then the homework is okay!

Week1/arrays.js
heartwhite: Understand. it's done. Thank you

marcianoviereck (approved):
Nice! The homework is approved:)

wouterkleijn:
Dear @heartwhite Congrats on the homework and getting your PR approved!  We really appreciate your response to the feedback of @marcianoviereck - super professional! Keep up the good work.

heartwhite:
Thank you Wouter. I will continue trying my best.

PR #104, opened 2019-04-29
Suheyb

suh3yb:
Javascript1 Week2 Homework

### DIFF
--- a/Week2/Homework/task10.js
+++ b/Week2/Homework/task10.js
@@ -24,4 +24,4 @@ if (m === n) {
     console.log("6/0 = 10/0");
 } else {
     console.log("An error occured!");
-};
+}
--- a/Week2/Homework/task10.js
+++ b/Week2/Homework/task10.js
@@ -9,7 +9,7 @@
 // in which you show that you understand the concepts (just 
 // like you've done in the above assignments).
 
-let multipleTypedArray = [0, "string", "000", true];
+const multipleTypedArray = [0, "string", "000", true];
 console.log("the type of my array is: " + typeof multipleTypedArray);
 console.log("the type of the first element of my array is: " + typeof multipleTypedArray[0]);
 console.log("the type of the second element of my array is: " + typeof multipleTypedArray[1]);
--- a/Week2/Homework/task10.js
+++ b/Week2/Homework/task10.js
@@ -1,0 +1,27 @@
+// TASK-10
+
+// 10. Write a program to answer the following questions:
+// 10.1 Can you store multiple types in an array? Numbers
+// and strings? Make an example that illustrates your answer.
+// 10.2 Can you compare infinities? (Not in Eyad's world)
+// - does 6/0 === 10/0? How can you test this?
+// 10.3 Add console.log statements to the above program 
+// in which you show that you understand the concepts (just 
+// like you've done in the above assignments).
+
+let multipleTypedArray = [0, "string", "000", true];
+console.log("the type of my array is: " + typeof multipleTypedArray);
+console.log("the type of the first element of my array is: " + typeof multipleTypedArray[0]);
+console.log("the type of the second element of my array is: " + typeof multipleTypedArray[1]);
+console.log("the type of the third element of my array is: " + typeof multipleTypedArray[2]);
+console.log("the type of the fourth element of my array is: " + typeof multipleTypedArray[3]);
+
+let m = 6 / 0;
+let n = 10 / 0;
+console.log("6/0 = " + m, "\n10/0 = " + n);
+
+if (m === n) {
+    console.log("6/0 = 10/0");
+} else {
+    console.log("An error occured!");
+};
--- a/Week2/Homework/task5.js
+++ b/Week2/Homework/task5.js
@@ -1,0 +1,21 @@
+// TASK 5
+
+/*
+5. How do you round the number 7.25, to the nearest integer (i.e., whole number)?
+5.1 Declare a variable z and assign the number 7.25 to it.
+5.2 console.log z.
+5.3 Declare another variable a that has the value of z but rounded to the nearest integer.
+5.4 console.log a.
+5.5 So now we have z and a find a way to compare the two values and store the 
+highest of the two in a new variable.
+5.6 console.log the highest value.
+*/
+
+let z = 7.25;
+console.log(z);
+
+let a = Math.round(z);
+console.log(a);
+
+let highestValue = Math.max(a, z);
+console.log(highestValue);
--- a/Week2/Homework/task8.js
+++ b/Week2/Homework/task8.js
@@ -17,12 +17,12 @@
 // 8.6 Make sure to also show a message when the 
 // variables you are comparing are not the same type.
 
-let foo = 9;
-let waldo = "find me";
-let awesome = true;
-let cities = ["Amsterdam", "Rotterdam", "Groningen"];
+const foo = 9;
+const waldo = "find me";
+const awesome = true;
+const cities = ["Amsterdam", "Rotterdam", "Groningen"];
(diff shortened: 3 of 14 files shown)


Week2/Homework/task10.js
```diff
@@ -0,0 +1,27 @@
...
+// 10. Write a program to answer the following questions:
+// 10.1 Can you store multiple types in an array? Numbers
+// and strings? Make an example that illustrates your answer.
+// 10.2 Can you compare infinities? (Not in Eyad's world)
+// - does 6/0 === 10/0? How can you test this?
+// 10.3 Add console.log statements to the above program 
+// in which you show that you understand the concepts (just 
+// like you've done in the above assignments).
+
+let multipleTypedArray = [0, "string", "000", true];
```
marcianoviereck: you can make this `let` a `const` because it's value won't change

Week2/Homework/task5.js
```diff
@@ -0,0 +1,21 @@
...
+5.6 console.log the highest value.
+*/
+
+let z = 7.25;
+console.log(z);
+
+let a = Math.round(z);
+console.log(a);
+
+let highestValue = Math.max(a, z);
```
marcianoviereck, line 20: nice! Good solution

Week2/Homework/task10.js
marcianoviereck: (And this applies to more parts of your code)

Week2/Homework/task8.js
```diff
@@ -0,0 +1,137 @@
...
+    console.log("\'" + cities + "\'" +" and "+ "\'" + car + "\'" + " HAVE the same type!");
+} else {
+    console.log("\'" + cities + "\'" +" and "+ "\'" + car + "\'" + " do NOT have the same type!");
+}
+
+if (typeof boo === typeof car) {
+    console.log("\'" + boo + "\'" +" and "+ "\'" + car + "\'" + " HAVE the same type!");
+} else {
+    console.log("\'" + boo + "\'" +" and "+ "\'" + car + "\'" + " do NOT have the same type!");
+}
```
marcianoviereck, line 136: Great!

marcianoviereck (changes requested):
The code looks really good! Just some small remarks here and there

Week2/Homework/task10.js
suh3yb: Thank you for your comments, I've made some changes as you requested.

suh3yb:
Thank you for your comments, I've made some changes as you requested.

marcianoviereck (approved):
Nice, homework is approved!

Week3/Homework/task3_morejs.js
```diff
@@ -0,0 +1,260 @@
...
+
+specialVehicle("yellow", 3, 2);
+
+// 10. Use the list of vehicles to write an advertisement. So that it prints something 
+//     like: "Amazing Joe's Garage, we service cars, motorbikes, caravans and bikes.". 
+//     (Hint: use a for loop.)
+
+// using join()
+function advertiseWithJoin() {
+    const vehicleList = vehicles.join("s, ");
```
marcianoviereck, line 114: nice solution!

Week3/Homework/task3_morejs.js
```diff
@@ -0,0 +1,260 @@
...
+// using for loop
+function advertiseWithLoop() {
+    let advertisementText = "Amazing Joe's Garage, we service ";
+    for (let e = 0; e < vehicles.length; e++) {
+        if (e !== vehicles.length - 1) {
+            advertisementText += vehicles[e] + "s, ";
+        } else {
+            advertisementText += vehicles[e] + "s.";
+        }
+    }
```
marcianoviereck, line 129: Good solution, you can also do it like this, but it's more complex so you don't have to change your code:) :

```
  for (let e = 0; e < vehicles.length; e++) {
    advertisementText += vehicles[e] + "s";
    advertisementText += (e === vehicles.length - 1) ? "." : ",";
   }
```

Week3/Homework/task3_morejs.js
```diff
@@ -0,0 +1,260 @@
...
+let o1 = { foo: "bar" };
+let o2 = { foo: "bar" };
+let o3 = o2;
+
+console.log(o3); // same as o2
+console.log("o3 === o2 : " + (o3 === o2)); // true
+console.log("o2 === o3 : " + (o2 === o3)); // true
+o2 = { boo : "far" };
+console.log(o2); // new o2
+console.log(o3); // does not change
```
marcianoviereck, line 223: here `o3` does not change because `o2` is `re-assigned` a value. So the whole object that `o2` was, is replaced by a new object, but that does not affect `o3`

Week3/Homework/task3_morejs.js
```diff
@@ -0,0 +1,260 @@
...
+console.log("o2 === o3 : " + (o2 === o3)); //false
+
+o2 = o3;
+console.log(o2); // {foo: "bar"}
+console.log("o3 === o2 : " + (o3 === o2)); // true
+o2.boo = "far";
+console.log(o2); // {foo: "bar", boo: "far"}
+console.log(o3); // {foo: "bar", boo: "far"} -- o3 changes but why now? is it because of order?
```
marcianoviereck, line 231: Ah yes, `o3` does change here because before (on line 216) you did `o3 = o2` and then later you did `o2 = o3`, therefore `o2` and `o3` are linked to eachother (they refer to the same location in memory). If a property of one changes, the other changes as well. (and the other way around). 

And so then when you change a property of `o2` (like you did with `o2.boo="far"` then `o3` will also change. 

[This](http://adripofjavascript.com/blog/drips/object-equality-in-javascript.html) explains it better I think 

I hope this makes it a bit more clear, it is a complicated subject!
Here is some more information on this: https://hackernoon.com/javascript-reference-and-copy-variables-b0103074fdf0 :)

Week3/Homework/task3_morejs.js
```diff
@@ -0,0 +1,260 @@
...
+console.log("o3 === o2 : " + (o3 === o2)); // true
+o2 = { foo: "bar" };
```
marcianoviereck, line 233: And then here you `re-assign` `o2`, which breaks the previous `link` of `o2 = o3`, so then if you would change `o3`, `o2` would not change with it.

Week3/Homework/task3_morejs.js
```diff
@@ -0,0 +1,260 @@
...
+console.log(o2); // { foo: "bar" }
+console.log(o3); // {foo: "bar", boo: "far"}
+console.log("o3 === o2 : " + (o3 === o2)); // false
```
marcianoviereck, line 236: Here is some more example code which might make this more clear:

```
let o1 = { foo: "bar" };
let o2 = { foo: "bar" };
let o3 = o2;

console.log(o1, o2, o3);
o2 = { boo : "far" };
console.log("1.", o1, o2, o3);

o2 = o3;
console.log("2.", o1, o2, o3);

o3.boo = "far";
console.log("3.", o1, o2, o3);

o2.boo = "car";
console.log("4.", o1, o2, o3);

o2 = { boo : "nar" };
console.log("5.", o1, o2, o3);

```

marcianoviereck (approved):
Nicely done! I've only added comments for explanation or alternative solutions, but the homework is approved so don't worry!

Week3/Homework/task3_morejs.js
suh3yb: nice one indeed, thanks!

suh3yb:
thank you very much for your time and feedback!

wouterkleijn:
Thanks @suh3yb for addressing Marciano's feedback!

PR #105, opened 2019-04-29
Hamza week2 homework

lizozo:
Included index.html file to see the results in console.


Week1/helloWorld.js
```diff
@@ -0,0 +1,12 @@
+//Turkish
+console.log('Merhaba Dünya');
+//English
+console.log('Hello World');
+//Spanish
+console.log('Hola Mundo');
+//French
+console.log('Bonjour Mond');
+//dutch
+console.log('Halo Werld');
```
marcianoviereck: Almost!:)

"Hallo Wereld":)

Week1/modulo.js
```diff
@@ -0,0 +1,10 @@
+let m = 7;
+m = m%3;
+console.log("When we divide 7 by 3, reminder is "+ m);
+let r = 458;
+let p = 23;
+let modulo = r%p;
+console.log("when you divide the number "+r+" by the number "+p+" the reminder is the number "+modulo);
+let two = 2;
+let reminder=r%two;
```
marcianoviereck: I would suggest adding some spaces to make the code more readable:

`let reminder = r % two;`

Week1/modulo.js
```diff
@@ -0,0 +1,10 @@
+let m = 7;
+m = m%3;
+console.log("When we divide 7 by 3, reminder is "+ m);
+let r = 458;
+let p = 23;
+let modulo = r%p;
+console.log("when you divide the number "+r+" by the number "+p+" the reminder is the number "+modulo);
+let two = 2;
```
marcianoviereck: You can also use `const` here, because the value of `two` won't change:)

Week1/typeCheck.js
```diff
@@ -0,0 +1,44 @@
...
+else {
+    console.log('DIFFERENT TYPE')
+};
+
+if (typeof myNumber == typeof myBoolean) {
+    console.log('SAME TYPE');
+}
+else {
+    console.log('DIFFERENT TYPE')
+};
```
marcianoviereck, line 43: I think you forgot to compare some types, such as `myArray` with `myNumber`

Week1/array.js
```diff
@@ -0,0 +1,11 @@
+let elements = [];
```
marcianoviereck: this could be `const` because the value won't change

marcianoviereck (changes requested):
The code looks good! Just some small things! Keep it up!

Week1/helloWorld.js
lizozo: :) corrected!

Week1/typeCheck.js
lizozo: Hello Marciano;

First of all thank you very much for your comments and feedbacks. I really appreciate it! Homework commands to create different types of variables so It was already obvious that all the result will be "DIFFERENT TYPE". so I thought it would be enough to test just some of them :)

Thank you very much again! Have a nice week

marcianoviereck (approved):
Nice! The homework is approved!:)

Week1/typeCheck.js
wouterkleijn: Thanks @lizozo for implementing the changes suggested by @marcianoviereck Also great to see that you are responding to his feedback - very professional!

PR #106, opened 2019-04-30
Abdulbawab

Albawab:


js1-2-1.js
```diff
@@ -0,0 +1,8 @@
+let Arabic = "مرحبا ايها العالم",
```
marcianoviereck, line 1: Fancy! This is also a nice way to do this!

js1-2-10.js
```diff
@@ -0,0 +1,19 @@
+let  multipleTypes = [2, "Sam", true];
```
marcianoviereck, line 1: This can be `const` since the value won't change anymore

js1-2-3.js
```diff
@@ -0,0 +1,17 @@
...
+console.log("the value of my variable x will be: undefined");
+
+// the value of x
+console.log(typeof(x));
+
+// add an value 
+  x = 5;
+
+  // the value of x
+  console.log("the value of x will be: number");
```
marcianoviereck: I don't think the value will be `number`:)

js1-2-5.js
```diff
@@ -0,0 +1,14 @@
...
+ let z = 7.25;
+ console.log(z);
+
+ let a = Math.round(z);
+ console.log(a);
+
+ // compare the two values
+ //Math.max(z,a);
+
+ console.log(Math.max(z,a));
```
marcianoviereck, line 14: Ah nice! Good solution!

js1-2-6.js
```diff
@@ -0,0 +1,26 @@
+// Declare an empty array
+let moreAnimals = [];
+let more_Sports = new Array();
```
marcianoviereck: You can remove this one, since you don't use it

js1-2-8.js
```diff
@@ -0,0 +1,19 @@
+let firstNumber = 7,
+    secondNumber = 10;
+
+console.log("The value of my variable foo is: " + firstNumber);
```
marcianoviereck: foo = firstNumber

it's fine, just that you wrote `foo` instead of `firstNumber`:)

js1-2-8.js
```diff
@@ -0,0 +1,19 @@
+let firstNumber = 7,
+    secondNumber = 10;
```
marcianoviereck: I think you can remove`secondNumber` because it is not used

js1-2-8.js
```diff
@@ -0,0 +1,19 @@
...
+console.log("The value of my variable foo is: " + firstNumber);
+console.log(typeof(firstNumber));
+
+if(typeof(firstNumber) == typeof(secondNumber)){
+    console.log("SAME TYPE");
+}
+
+let firstString = "10";
+
+if(typeof(firstNumber) == typeof(firstString)){
```
marcianoviereck: I think you forgot to add more variable (with different types):

```
8.1 First declare at least four variables and assign them different data types.
```

marcianoviereck (changes requested):
The homework looks good! and nice!

I think you forgot a small part of the homework though

js1-2-8.js
```diff
@@ -0,0 +1,24 @@
...
+    console.log("SAME TYPE");
+}
+
+
+/* let firstNumber = 7,
+secondNumber = 10; */
+
+let firstString = "10",
+    first = 0;
+
```
marcianoviereck, line 17: Ah, I think it's almost correct, I think you need to add two variables:

one for the `array` data-type and one of the `boolean` data type. Then compare those as well. Then the homework is approved!

marcianoviereck (changes requested):
One last thing, I think you forgot two data-types in the `js-1-2-8.js` file

wouterkleijn:
Thanks @marcianoviereck for providing feedback. @Albawab Can you please address Marciano's comments? Thanks!

js1-2-8.js
```diff
@@ -12,8 +12,8 @@ if(typeof(firstNumber) == typeof(secondNumber)){
 /* let firstNumber = 7,
 secondNumber = 10; */
 
-let firstString = "10",
-    first = 0;
+let firstString = ture,
```
marcianoviereck, line 15: this is not valid a boolean value (and will not compile). If you can fix this then it is fine

marcianoviereck (changes requested):
I found one last thing that is not correct yet.

PR #108, opened 2019-05-01
hani's homework

hanialkudsi:


js.js
```diff
@@ -0,0 +1,124 @@
...
+console.log(z);
+let a = Math.floor(z);
+console.log(a);
+let h;
+if (z>a){
+    h = z ;
+}else {
+    h = a;
+};
+console.log("the highest value is "+h);
```
marcianoviereck, line 35: The if else statement is nice! You can also use [Math.max](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max) which does the same thing:)

js.js
```diff
@@ -0,0 +1,124 @@
...
+console.log("the value of y now is 'new string'");
+console.log(y);
+
+
+let z = 7.25;
+console.log(z);
+let a = Math.floor(z);
+console.log(a);
+let h;
+if (z>a){
```
marcianoviereck, line 30: If you add some spaces here, it becomes easier to read:

```
if (z > a) {
```

js.js
```diff
@@ -0,0 +1,124 @@
...
+    h = a;
+};
+console.log("the highest value is "+h);
+
+let friends = [];
+console.log("the value is array");
+console.log(friends);
+let myFavoriteAnimals = ["cat's","fish's","turtle's","bird's"];
+console.log(myFavoriteAnimals);
+myFavoriteAnimals[4] = "Daan's";
```
marcianoviereck, line 42: Haha, this works! But a cleaner way to do this by using:

```
myFavoriteAnimals.push("Daan's");
```

js.js
```diff
@@ -0,0 +1,124 @@
...
+
+
+let myString = "this is a test";
+console.log(myString);
+console.log(myString.length);
+
+let m = 5;
+let n = "rami";
+let o = [1];
+let p = true;
```
marcianoviereck, line 53: If you change the naming of your variables, it describes more about what they are 

(So instead of `m` you can use `age`, etc)

js.js
```diff
@@ -0,0 +1,124 @@
...
+//multiple type array
+let myArray = [
+    "me",
+    2,
+    "you2",
+    4,
+    true
+];
+//comparing infinties
+if (6/0 === 10/0){
```
marcianoviereck, line 119: If you first store the results of `6/0` and `10/0` in variables, then it becomes easier to use them later on (and the code becomes more readable)

js.js
```diff
@@ -0,0 +1,124 @@
+console.log("hello world!");
```
marcianoviereck, line 1: Can you split up your homework in separate files (as was asked in the assignment)?

```
For all the following exercises create a new .js file. Try to find a proper name for each file or make a small comment about what it does inside for future reference.
```

marcianoviereck (changes requested):
The code looks really good! I only have some small remarks (but more on making the code cleaner and easier to read).

And if you can split your homework into separate files, then that would be great:)

js.js
hanialkudsi: yup ...I'm really sorry ! 
i didn't notice that..

js.js
hanialkudsi: thanx .I'll pay more attention for this details

js.js
hanialkudsi: okay ,i only use this to give a certain position to "daan's"

wouterkleijn:
Thanks for addressing @marcianoviereck 's comments. @marcianoviereck Is this PR finished? Also, @hanialkudsi it seems that you added your homework of the last week (week3) was added to the same PR as the week before.

marcianoviereck:
@wouterkleijn yeah the homework for week 2 is approved. But I still need to check the homework of week 3.. since it is also added to this PR.

new.js
```diff
@@ -0,0 +1,161 @@
...
+colorCar("yellow");
+
+//Q3
+//Create an object and a function that takes the object as a parameter and prints out all of its properties and values
+const movie = {
+    name: "zohan",
+    year: 2000,
+    genre: "comedy",
+    actor: "adam sandler"
+    };
```
marcianoviereck, line 51: If you can remove the extra `tab` here then it's easier to read:

So like this:

```
const movie = {
    name: "zohan",
    year: 2000,
    genre: "comedy",
    actor: "adam sandler"
};
```

new.js
```diff
@@ -0,0 +1,161 @@
...
+
+    function movieInfo(object){
+        let entries = Object.entries(object);
+        for ( let [key,value] of entries){
+    console.log(key + " : " + value);
```
marcianoviereck, line 56: If you add some extra tabs here, it's more readable:

```
    function movieInfo(object){
        let entries = Object.entries(object);
        for ( let [key,value] of entries){
           console.log(key + " : " + value);
        }  
    }
```

hanialkudsi:
i add it here first by mistake, but then i add it again on a new branch

new.js
```diff
@@ -0,0 +1,161 @@
...
+    
+      if (i == listOfVehicles.length-1){
+        adv +=  " and " + listOfVehicles[i] + ".";
+      }else if(i == listOfVehicles.length-2){
+         adv += listOfVehicles[i];
+      }
+      else { adv += listOfVehicles[i] +  ", " ;
+      }
+}console.log(adv)
+}
```
marcianoviereck, line 146: This is fine, but it could be a bit more readable if you add more spaces, like this:

```
function advertisement (){
    for( i = 0 ; i < listOfVehicles.length; i++) {
      if (i == listOfVehicles.length-1){
        adv +=  " and " + listOfVehicles[i] + ".";
      } else if(i == listOfVehicles.length-2) {
         adv += listOfVehicles[i];
      } else { 
         adv += listOfVehicles[i] +  ", " ;
      }
   }
  console.log(adv);
}
```

Then it is also easier to see which `{` matches to the `}`

new.js
```diff
@@ -0,0 +1,161 @@
...
+
+listOfVehicles.push("truck");
+console.log(listOfVehicles);
+advertisement();
+
+//Q12
+//create an empty object
+let emptyOpject = {};
+
+//Q13
```
marcianoviereck, line 160: I think you forgot question `14`, `15`, `16` and `17`, see the [homework](https://github.com/HackYourFuture/JavaScript1/blob/master/Week3/MAKEME.md)

marcianoviereck (changes requested):
The homework looks good, but I think you forgot some parts of the homework of week 3

hanialkudsi:
i just saw that, i write every thing actually but i think it weren't uploaded to github. any way it's still written on my version of the home work
Q13
const myTeachers = [
    {
      firstName: "Ze",
      lastName: "German",
    },
    {
      firstName: "Unmesh",
      lastName: "Joshi",
    },
    {
      firstName: "Sander",
      lastName: "Hidding",
    },
  ];

  //Q14
  //Add a property to the object you just created that contains the languages that they have taught you

  function addingLanguage (){
    for (i = 0 ; i < myTeachers.length; i++ ){
      if(myTeachers[i].lastName === "Joshi"){
       myTeachers[i].languages = "git"; 
      }else if (myTeachers[i].lastName === "Hidding"){
       myTeachers[i].languages = "java script"; 
      }else if (myTeachers[i].lastName === "German"){
       myTeachers[i].languages = "html & css"; 
      };
    }
  console.log(myTeachers);
}
addingLanguage();

//Q15
//Write some code to test two arrays for equality using == and ===

let x = [1, 2, 3];
let y = [1, 2, 3];
let z = y;

console.log(x == y ? true : false );
console.log(x === y ? true : false );
console.log(z == y ? true : false );
console.log(z == x ? true : false );

//Q16
//Show that changing o2 changes o3 (or not) and changing o1 changes o3(or not).

let o1 = { foo: "bar" };
let o2 = { foo: "bar" };
let o3 = o2;

function changingValues(){
  console.log("old value for o3 ", o3);
  o2.edit = "new key";
  console.log("new value for o3 after changing o2 ", o3);
  o1.edit = "amsterdam";
  console.log("the value of o3 after changing the value of o1 ", o3);
}
changingValues()
// changing o2 changes o3 but changing o1 dosen't effect o3 because they aren't realated to each other

//Q17
//What does the following code return? (And why?)

let bar = 42;
console.log(typeof typeof bar);

let x = true
console.log(typeof typeof x);
//by typing typeof twice we git the type of the result which is a string because it's written between quotations marks

marcianoviereck:
@hanialkudsi The code in your comment is good! But can you upload it through GIT? Then I will approve the PR!

PR #110, opened 2019-05-05
Riam Alali-week2hw

RiamAlali:


week2homework.js
```diff
@@ -0,0 +1,121 @@
...
+console.log (TurkishGreeting)
+let DutchGreeting = "Hallo Wereld!";
+console.log (DutchGreeting)
+
+//2
+console.log ('I\'m awesome');
+
+//3
+let x ;
+console.log ("the value of my variable x will be:EvenNumber");
```
marcianoviereck, line 18: Are you sure that the value of `x` is `EvenNUmber` ?
(I don't think it will be that value)

week2homework.js
```diff
@@ -0,0 +1,121 @@
...
+console.log (x);
+x = 8 ;
+console.log ("the value of my variable x will be:EvenNumber");
```
marcianoviereck, line 21: Are you sure that the value of x is `EvenNumber`?

week2homework.js
```diff
@@ -0,0 +1,121 @@
...
+console.log (x)
+
+//4
+let y = "HackYourFuture"
+console.log ("the value of the string is "+"HackYourFuture")
```
marcianoviereck, line 26: Please don't forget to add `;` at the end of the line:)

week2homework.js
```diff
@@ -0,0 +1,121 @@
...
+console.log (y)
+
+//5
+let z = 7.25;
+console.log(z);
+let a = (Math.round(z));
+console.log(a);
+
+let highest;
+if (z>a){
```
marcianoviereck, line 39: Good! You can also use [math.max](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max), but this solution is fine:)

week2homework.js
```diff
@@ -0,0 +1,121 @@
...
+    highest=z;
+}
+else{
+    highest=a;
+}
+console.log(highest);
+
+
+//6
+let FavoriteAnimals = [ ];   //6.1
```
marcianoviereck, line 49: You should not use capital letters for the variable names.

So `FavoriteAnimals` should be `favoriteAnimals`.

See [The naming conventions](https://github.com/HackYourFuture/fundamentals/blob/master/fundamentals/naming_conventions.md)

week2homework.js
```diff
@@ -0,0 +1,121 @@
...
+console.log("the value of my array is : Empty."); //6.2
+console.log(FavoriteAnimals); //6.3
+let MyFavoriteAnimals = [ 'Deer' , 'Horse' , 'Koala' ]; //6.4
+console.log (MyFavoriteAnimals);  //6.5
+let NewMyFavoriteAnimals = MyFavoriteAnimals.concat(['baby pig']); //6.6
```
marcianoviereck, line 54: Ah nice, you can also use 
`MyFavoriteAnimals.push('baby pig');`

Which might be an easier solution

week2homework.js
```diff
@@ -0,0 +1,121 @@
...
+let type3 = " Hello JavaScript " ;
+console.log ("The value of my variable type3 is: " + type3);
+console.log (typeof(type3));
+let type4 = "Class21" ;
+console.log ("The value of my variable type4 is: " + type4);
+console.log (typeof(type4));
+
+//8.5
+if (typeof(type1)===typeof(type2))
+{console.log('type1 and type2 are SameType');}
```
marcianoviereck, line 78: Tip: The code is easier to read if you add more spaces in between the if statement, like this:

```
if (typeof(type1)===typeof(type3)) {
  console.log('type1 and type3 are SameType');
} else {
  console.log('type1 and type3 are NOT SameType');
}
```

week2homework.js
```diff
@@ -0,0 +1,121 @@
...
+if (typeof(type3)===typeof(type4))
+{console.log('type3 and type4 are SameType');}
+else {console.log('type3 and type4 are NOT SameType');}
+
+
+
+
+
+//9
+let X = 7 ;
```
marcianoviereck, line 106: please use lowercase `x`

week2homework.js
```diff
@@ -0,0 +1,121 @@
+//1
```
marcianoviereck, line 1: I see that you did all your homework in one file, but the assignment was to make one file for each part of the homework:

marcianoviereck (changes requested):
Overall it looks good! Just some smaller style remarks and could you split your homework up in separate files?

week2homework.js
RiamAlali: Thank you for your review, I'v made changes as advised, could you please re-check?

wouterkleijn:
Dear @RiamAlali Thanks a lot for addressing Marciano's feedback - very professional indeed! @marcianoviereck Can you check if there is anything left to be done?

marcianoviereck:
@wouterkleijn Yes the homework for week 2 is now approved, but I still need to check the homework for week 3..

week2homework.js
marcianoviereck: Can you fix this one?

week2homework.js
marcianoviereck: And this one, then the homework is approved!

marcianoviereck (changes requested):
Some very minor changes left, over all its good just two small things!

PR #111, opened 2019-05-05
JS1 Homework/Week2

Shvan1:


app.js
```diff
@@ -0,0 +1,77 @@
+
+console.log("Hello, World!"); //->English
+console.log("Silav, Cihan!");//->Kurdish
+console.log("Hoi, Wereld!"); //->Dutch
+
+console.log("I'm awesome!");
+
+let x =
```
marcianoviereck: I don't think this will compile:) 

I think you wanted to do this?
`let x;` ?

app.js
```diff
@@ -0,0 +1,77 @@
+
+console.log("Hello, World!"); //->English
+console.log("Silav, Cihan!");//->Kurdish
+console.log("Hoi, Wereld!"); //->Dutch
+
+console.log("I'm awesome!");
+
+let x =
+console.log('I think the value of x is bigger than 5');
```
marcianoviereck: This is also not correct, because the value of `x` is not set (yet)

app.js
```diff
@@ -0,0 +1,77 @@
...
+console.log("Silav, Cihan!");//->Kurdish
+console.log("Hoi, Wereld!"); //->Dutch
+
+console.log("I'm awesome!");
+
+let x =
+console.log('I think the value of x is bigger than 5');
+console.log("The value of x is 6");
+let x = 6;
+console.log(I think the value of x is 6);
```
marcianoviereck: I think that this will will give an error message.


Did you mean to do this:

`console.log("I think the value of x is 6");`
?

app.js
```diff
@@ -0,0 +1,77 @@
...
+console.log(y);
+
+const z = 7.25;
+console.log(z);
+const a = z;
+console.log(a);
+var round =Math.round(7.25);
+console.log(round)
+
+let arrays =
```
marcianoviereck: Same here as above, I don't think this will compile

app.js
```diff
@@ -0,0 +1,77 @@
...
+console.log("I think variable arrays is related to animals");
```
marcianoviereck: But what do you think is inside the array? What is the value of the array?

app.js
```diff
@@ -0,0 +1,77 @@
...
+console.log(arrays);
+let myFavoriteAnimals = [cat, dog, sheep];
+console.log(myFavoriteAnimals);
+let favoriteAnimalOfDaan = "baby pig";
+const animals = "myFavoriteAnimals" + "" + "favoriteAnimalOfDaan";
```
marcianoviereck: Ah, this is not correct.

You're making a `string` with the value `myFavoriteAnimals` and then adding `favoriteAnimalOfDaan` to it.
So it will result in "myFavoriteAnimalsfavoriteAnimalOfDaan", instead of an `array` with the values in it.

app.js
```diff
@@ -0,0 +1,77 @@
...
+console.log(animals);
+
+let myString = "this is a test";
+console.log(myString);
+console.log(myString ["length"]);
```
marcianoviereck: myString.length is a better solution (I'm not even sure if `myString ["length"]` compiles)

app.js
```diff
@@ -0,0 +1,77 @@
...
+
+let num = 11;
+let address = "Huygenhoekring";
+let supermarkets = ["Lidl", "Dekka", "Jumbo"];
+let booleans = true, false, true;
+console.log(num);
+console.log(address);
+console.log(supermarkets);
+console.log(booleans);
+console.log(I think the type of num is number,
```
marcianoviereck: Did you forget tot add " " in the console log? (because I dont think this compiles)

app.js
```diff
@@ -0,0 +1,77 @@
...
+console.log(s % v);
+console.log(v % h);
+
+const mix = [5, "Shvan"];
+console.log(mix);
+
+var maxNumber = Math.pow(6/0);
+
+if (maxNumber === Infinity)
+console.log(0 / maxNumber);
```
marcianoviereck, line 140: I would recommend adding brackets here, so like this:
```
if (maxNumber === Infinity) {
  console.log(0 / maxNumber);
}
```

It makes it easier to avoid mistakes

marcianoviereck (changes requested):
In general the homework looks good, but there are some remarks here and there.

And can you split up the homework in separate files?

app.js
```diff
@@ -0,0 +1,77 @@
+
```
marcianoviereck, line 8: Can you split up the homework in separate files?

app.js
Shvan1: let x;

wouterkleijn:
@marcianoviereck I noticed @Shvan1  made some edits. Is this okay now or does he need to make some further changes? Thanks!

marcianoviereck:
@wouterkleijn not all the feedback is resolved yet, as far as I can see in the 'files changed'. There are still some remarks of me which have not been fixed yet.

marcianoviereck (changes requested):
Not all my comments have been adressed yet, as far as I can see. Can you still make those changes?

app.js
Shvan1: It is solved

app.js
Shvan1: changed to : "I think the value of x is undefined"

app.js
Shvan1: Changed to : let arrays;

app.js
Shvan1: Changed to: "I think the value will be undefined"

app.js
Shvan1: Changed to: const animals = myFavoriteAnimals + "" + favoriteAnimalOfDaan;

app.js
Shvan1: Changed to: console.log("the value of x: 6");

app.js
Shvan1: Changed to: console.log(myString.length);

app.js
Shvan1: Changed to: console.log("I think the type of num is number, type of address is string, type of supermarkets is array and type of booleans is boolean.");

app.js
Shvan1: Changed to: var maxNumber = Math.pow(6/0);

if (maxNumber === Infinity){
console.log(0 / maxNumber);
}

app.js
```diff
@@ -1,80 +1,142 @@
...
+//6.4
 let myFavoriteAnimals = [cat, dog, sheep];
+
+//6.5
 console.log(myFavoriteAnimals);
+
+//6.6
 let favoriteAnimalOfDaan = "baby pig";
-const animals = "myFavoriteAnimals" + "" + "favoriteAnimalOfDaan";
+const animals = myFavoriteAnimals + "" + favoriteAnimalOfDaan;
```
marcianoviereck: Ah this one is still incorrect,you should use `myFavoriteAnimals.push('baby pig')`. 
Read more about it here: https://developer.mozilla.org/nl/docs/Web/JavaScript/Reference/Global_Objects/Array/push

marcianoviereck (changes requested):
One last change request, for the rest the homework looks good!
If you fix that last one, then the homework is approved.

app.js
Shvan1: Changed to: const animals = myFavoriteAnimals.push('baby pig');
Just last week, we have covered the .push() method.

marcianoviereck (approved):
Ah nice, thanks for fixing the homework! It's approved now :)

PR #114, opened 2019-05-09
Abdullah Al Ahmad

AbduCodecamp:
JavaScript homework - week1


Week1/js1-w1-A.A.A.js
```diff
@@ -0,0 +1,175 @@
...
+
+// Debug single quote issue
+
+console.log('I\'m awesome');
+
+// Variables & Values
+
+var x;
+
+console.log("the value of my variable x will be: a number");
```
marcianoviereck: I don't think the value of `x` will be `a number`, are you sure this is correct?

Week1/js1-w1-A.A.A.js
```diff
@@ -0,0 +1,175 @@
...
+
+console.log(x);
+
+x = 12;
+
+console.log("the value of my variable x will be: a number");
```
marcianoviereck: I don't think the value of `x` will be `a number`, are you sure this is correct?

Week1/js1-w1-A.A.A.js
```diff
@@ -0,0 +1,175 @@
...
+
+console.log(x);
+
+var y = 'Tom';
+
+console.log("the value of my atring y will be: a name");
```
marcianoviereck: Here the same:) as above

Week1/js1-w1-A.A.A.js
```diff
@@ -0,0 +1,175 @@
...
+
+var z = 7.25;
+
+console.log(z);
+
+// Round z to the closest integer
+
+var round = Math.round;
+
+var a = round(z);
```
marcianoviereck: Ah, you can also do this in one line, like this:

`const a = Math.round(z);`

Week1/js1-w1-A.A.A.js
```diff
@@ -0,0 +1,175 @@
...
+console.log ('Hola Mundo!'); // Spanish
+
+
+// Debug single quote issue
+
+console.log('I\'m awesome');
+
+// Variables & Values
+
+var x;
```
marcianoviereck: Don't use `var` please, use `let` or `const`

Week1/js1-w1-A.A.A.js
```diff
@@ -0,0 +1,175 @@
...
+
+var a = round(z);
+
+console.log(a);
+
+// Print higher value z or a 
+
+if (a<z) {
+    var i = z;
+    console.log(i);
```
marcianoviereck: you can also do this directly:

```
if (a<z) {
    console.log(z);
} else {
    console.log(a);
}
```

marcianoviereck: Or with [Math.max](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max)

Week1/js1-w1-A.A.A.js
```diff
@@ -0,0 +1,175 @@
...
+    var i = a;
+    console.log(i);
+
+}
+
+// Animals Names
+
+var myFavoriteAnimals;
+
+console.log('The value of heros is: animales\'s names');
```
marcianoviereck: I don't think this is correct, `heros` is not the name of the variable and the value is incorrect

Week1/js1-w1-A.A.A.js
```diff
@@ -0,0 +1,175 @@
...
+
+console.log(myFavoriteAnimals);
+
+myFavoriteAnimals = ['monkey', 'horse', 'turtle'];
+
+console.log(myFavoriteAnimals);
+
+myFavoriteAnimals = myFavoriteAnimals + [', baby pig'];
```
marcianoviereck: A cleaner (and better) solution is:

`myFavoriteAnimals.push('baby pig');`

I think this works better:)

Week1/js1-w1-A.A.A.js
```diff
@@ -0,0 +1,175 @@
+// Greetings
```
marcianoviereck: Can you split your homework up in different files? As requested in the assignment:

marcianoviereck (changes requested):
Most of the homework looks fine.

Try to avoid using `var`, but use `let` or `const` instead. And also split up the homework in separate files.

wouterkleijn:
Thanks @AbduCodecamp for addressing some of Marciano's feedback. Next time also place some comments directly answering his concerns if possible. @marcianoviereck Can you have a final look? Thanks!

Week2/arrays.js
```diff
@@ -0,0 +1,15 @@
...
+
+console.log('The value of myFavoriteAnimals is: undefined then an object \'array\' ');
+
+console.log(myFavoriteAnimals);
+
+myFavoriteAnimals = ['monkey', 'horse', 'turtle'];
+
+console.log(myFavoriteAnimals);
+
+myFavoriteAnimals = myFavoriteAnimals + [', baby pig'];
```
marcianoviereck, line 13: It's better to use `myFavoriteAnimals.push('baby pig')'`. This will work but it's not the correct way to do this.

marcianoviereck (changes requested):
Looks good, I just have one last minor comment, if that is fixed then the homework of week 2 is approved. I will still check the week 3 homework

Week3/advertisement.js
```diff
@@ -0,0 +1,26 @@
...
+
+      if (i == vehicles.length-1){
+        ad += 's' + ' and ' + vehicles[i] + 's' + '.';
+      }else if(i == vehicles.length-2){
+         ad += vehicles[i];
+      }
+      else { ad += vehicles[i]+ 's' +  ', ' ;
+      }
+}console.log(ad)
+}
```
marcianoviereck, line 18: This is fine, it would be more readable if you add more spaces and align your brackets (`{`), like this:

And I would advise to put the code of `'s' + ' and '` not in that part, but always add the `s` in the other if statement. It's difficult to explain so here is the code of what I mean:

```
let ad = "Amazing Joe's Garage, we service ";
function advertisement (){
    for( i = 0 ; i < vehicles.length; i++){
      if (i == vehicles.length-1) {
        ad += ' and ' + vehicles[i] + 's' + '.';
      } else if(i == vehicles.length-2) {
         ad += vehicles[i]; + 's' 
      } else { 
        ad += vehicles[i]+ 's' +  ', ' ;
      }
   }
 console.log(ad);
}
```

Week3/sumFun.js
```diff
@@ -0,0 +1,13 @@
+// Create a function that takes 3 arguments and returns the sum of the these arguments.
+
+function sum(){
+    let sum =0; 
+    for(let i=0;i<arguments.length;i++){
```
marcianoviereck, line 5: technically this is correct, but I think that the assignment meant that you explicitly define three arguments and combine those. but I guess you know how to do that, since you've done it in a more complex way. So it's okay

marcianoviereck: A tip: This is a pretty clever solution actually. But more difficult in use for other programmers if you collaborate on the code, because it is not intuitive for someone to know how many arguments they have to pass to the function and what will happen to them. So this is just a tip for the future, this solution is clever but not really good if you are collaborating with other programmers.

Week3/vehicle.js
```diff
@@ -0,0 +1,18 @@
+// Create a function called vehicle, like before, but takes another parameter called age, so that vehicle("blue", 1, 5) prints 'a blue used car'
+
+function vehicleType(color,code, age){
```
marcianoviereck, line 3: This is also correct, but it's not a good practice to modify the arguments that are passed in the function. Because if you modify a property of that argument in the function,  it will also change the property for that argument outside the function. I know this is not the case in this specific function, but it could happen in the future, which could lead to confusing results. For example:

```
function vehicleType(color,code, anObject){
    if (anObject.age < 1) {
        anObject.age = 'bused'
    } else {
       anObject.age = ' new'
    }

    if (code == 1){
        console.log('a ' + color + anObject.age + ' car');
    } else if (code == 2) {
        console.log('a ' + color + anObject.age + ' motorbike');
    }
}

let color = "green";
let code = 1;
let anObject = { age: 12, name: "test" };

console.log("before the function, the age is " + anObject.age);
vehicleType(color, code, anObject);
console.log("after the first function, the age is " + anObject.age);
```

In this code example, the property `age` is modified in the function, which affects the `anObject` variable, which is outside the function.

Week3/vehicleList.js
```diff
@@ -0,0 +1,28 @@
+// Make a list of vehicles, you can add "motorbike", "caravan", "bike", or more.
+
+// How do you get the third element from that list?
+
+// Change the function vehicle to use the list of question 7. So that vehicle("green", 3, 1) prints "a green new bike".
+
+ let vehicles = [' car', ' motorbike', ' bike', ' caraven' ];
+
+function vehicleType(color,code, age){
+    if (age < 1) {
```
marcianoviereck, line 10: here the same remark as above (about the bad practice to modify an argument that is passed in the function)

Week3/index.html
```diff
@@ -0,0 +1,34 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>JS-week3</title>
+</head>
+<body 
+style ="text-align: center; padding: 10vw;">
+
+    <h1 style="color: red; font: 3em sans-serif;">JS-week3</h1>
+    
+    <script src="strings.js"></script>`
```
marcianoviereck, line 14: I think you forgot some parts of the homework?

I think assignment 12 untill 16 are missing, see the [homework](https://github.com/HackYourFuture/JavaScript1/blob/master/Week3/MAKEME.md)

marcianoviereck (changes requested):
Homework for week 2 is approved, but for week 3 I have some remarks. And I think that you forgot some parts of the homework (see my other comments)

PR #116, opened 2019-05-14
Abdulbawab js1 week3

Albawab:


function.js
```diff
@@ -0,0 +1,105 @@
...
+    old : 30
+}
+
+let variable = Object.keys(obje);
+let value = Object.values(obje);
+
+for(let i = 0; i < variable.length; i++){
+    console.log(obj(variable[i]));
+    console.log(obj(value[i]));
+}
```
marcianoviereck, line 36: Good solution, just a tip, there are more [ways](https://zellwk.com/blog/looping-through-js-objects/) to do this!

function.js
```diff
@@ -0,0 +1,105 @@
...
+vehicleType("blue", 2)
+
+// 5 without the if
+3 === 3 ? console.log("yes") : console.log("no");
+
+// 6 a function called vehicle with parameter called age
+function vehicle(color , code , age){
+    if(code == 1)
+    console.log("a "+ color + " car");
+
```
marcianoviereck: Ah, I think you forgot the `}`

I would also recommend to always add brackets in your `if-statement` (even if it is only one line). It helps avoid silly mistakes:

```
if(code == 1) {
    console.log("a "+ color + " car");
}
```

function.js
```diff
@@ -0,0 +1,105 @@
...
+let advertisement = "\"Amazing Joe's Garage, we service ";
+var loop = "";
+for(let i = 0; i <vehicles.length-1; i++){
+   loop += vehicles[i] + ", ";
+}
+if(vehicles[3]){
+    loop = loop +" and " + vehicles[3]+"\"";    
+ }
+console.log(advertisement + loop);
+}
```
marcianoviereck: I think that this `}` is not supposed to be here?

function.js
```diff
@@ -0,0 +1,105 @@
...
+//prints "a green new bike".
+console.log("a "+ color +" " +vehicles[code - 1]);
+
+// 10 write an advertisement
+let advertisement = "\"Amazing Joe's Garage, we service ";
+var loop = "";
+for(let i = 0; i <vehicles.length-1; i++){
+   loop += vehicles[i] + ", ";
+}
+if(vehicles[3]){
```
marcianoviereck: What if we add one more `vehicle` to the list of `vehicles`, does this still work?
I would recommend changing this into a function, and to make it work regardless of the amount of `vehicles` in the list.

function.js
```diff
@@ -0,0 +1,105 @@
...
+
+vehicle("blue", 1, 5);
+vehicle("green", 3, 5);
+
+// 12 Create an empty object.
+let emtyObject= new Object;
+
+// 13 teachers 14
+let teacher = {
+    Amen : "history",
```
marcianoviereck: I think step 14 needs to be done separately. |

`Add a property to the object you just created that contains the languages that they have taught you.`

Now the language is not added as a `property` but as a `value`. Can you change that?

function.js
```diff
@@ -0,0 +1,105 @@
...
+
+console.log(x==y); //false
+console.log(x===y);//false
+console.log(z==y);//true
+console.log(z==x);//false
+
+//Does the order that you assign (o3 = o2 or o2 = o3) matter? Yes
+
+let bar = 42;
+typeof typeof bar;
```
marcianoviereck, line 119: I think that you forgot assignment 16 and 17 (or at least, forgot to finish them)?

marcianoviereck (changes requested):
Most of the homework looks good! But there are some remarks here and there and I think that you forgot the last two assignments!

function.js
Albawab: I did that.

function.js
Albawab: done

function.js
```diff
@@ -0,0 +1,106 @@
...
+
+
+vehicle("blue", 1, 5);
+vehicle("green", 3, 5);
+
+// 12 Create an empty object.
+let emtyObject= new Object;
+
+// 13 teachers 14
+let teacher = {
```
Albawab, line 83: Done

function.js
Albawab: @marcianoviereck  Done

Albawab: // 16

let o1 = { foo: "bar" };
let o2 = { foo: "bar" };
let o3 = o2;

//Show that changing o2 changes o3 (or not) // YES changes
//changing o1 changes o3(or not).// NO changes
// Does the order that you assign (o3 = o2 or o2 = o3) matter? It is matter



// 17
let bar = 42;
typeof typeof bar;
 // It is String first i have call the type of the value . second the type of the value was number and the word number is string

Albawab: @marcianoviereck  i have alredy add 16 and 17

function.js
marcianoviereck: I don't think that this is correct. The `advertisement` should be a function. And this code doesn't work anymore if  I add more (or less) values to the `vehicles` array.

For example, if `vehicles` look like this:
`let vehicles = ["motorbike", "caravan", "bike", "car", "train", "airplane"];`

Then your code will not work anymore right?

So it is almost correct, but quite.

Do you understand what I mean?

function.js
```diff
@@ -0,0 +1,117 @@
...
+console.log(z==y);//true
+console.log(z==x);//false
+
+// 16
+
+let o1 = { foo: "bar" };
+let o2 = { foo: "bar" };
+let o3 = o2;
+
+//Show that changing o2 changes o3 (or not) // YES changes
```
marcianoviereck: I understand that you know that it does (or does not) change, but can you show it? With some examples and some `console.log`s?

marcianoviereck (changes requested):
Most of the homework looks good! But there are 2 (new) comments, if you fix those then the homework is approved!

function.js
Albawab: Done

marcianoviereck (approved):
Thanks for fixing the homework! It's approved now :)

PR #120, opened 2019-05-14
Abdullah week3 JS homework

heartwhite:
!!! important..!!  
here there are some changes from old weeks . please look after "Week3/arrays.js"
 12 changes on head are old. (because I did my changes on pull requests page according to Marcianos feedback. afterwards I applied changes on my page on forked repo. that's why the files seem changed ... I'm sorry for the mess...

JUST klick "WEEK3"  (the link one before the last link) for HOMEWORK

### DIFF
--- a/Week1/numberRounding.js
+++ b/Week1/numberRounding.js
@@ -1,0 +1,10 @@
+//round and compair numbers
+let z = 7.25;
+console.log(z);
+let a = Math.round(z);
+console.log(a);
+let numbers = [a, z];
+numbers.sort(function(a, b) { return b - a; });
+
+var highest = numbers[0];
+console.log(highest);
--- a/Week1/numberRounding.js
+++ b/Week1/numberRounding.js
@@ -4,7 +4,12 @@ console.log(z);
 let a = Math.round(z);
 console.log(a);
 let numbers = [a, z];
-numbers.sort(function(a, b) { return b - a; });
+let highNumber=(Math.max(z,a));
+console.log(highNumber);
 
+
+/* ----this is what i found first---
+numbers.sort(function(a, b) { return b - a; });
 var highest = numbers[0];
-console.log(highest);
+console.log(highest); 
+*/
--- a/Week3/moreJs-10.js
+++ b/Week3/moreJs-10.js
@@ -1,0 +1,26 @@
+/*
+Use the list of vehicles to write an advertisement.
+ So that it prints something like:
+"Amazing Joe's Garage, we service cars,
+motorbikes, caravans and bikes.". 
+(Hint: use a for loop.)
+*/
+function advertisement(){
+    // let vehicles = [' motorbike' , ' caravan', ' bike', ' car'];
+    // variable 'vehicles' is can be read from  moreJs-7and8.js
+    // but i used it again inside of function to show you,
+    // and in case you can run in a html seperately.
+    let products = [];
+    for(i=0; i<vehicles.length; i++){
+        products.push(' ' + vehicles[i]+'s');
+    }
+    let lastTwoProduct = products[products.length - 2] + ' and' + products[products.length - 1] ;
+    console.log(lastTwoProduct);
+    products.splice((products.length - 2), 2,);
+
+    console.log('Amazing Abdullah\'s Garage, we sell ' + products.join() + ',' + lastTwoProduct +'.');
+    
+
+}
+advertisement();
+
--- a/Week3/moreJs-10.js
+++ b/Week3/moreJs-10.js
@@ -5,22 +5,19 @@ Use the list of vehicles to write an advertisement.
 motorbikes, caravans and bikes.". 
 (Hint: use a for loop.)
 */
-function advertisement(){
-    // let vehicles = [' motorbike' , ' caravan', ' bike', ' car'];
-    // variable 'vehicles' is can be read from  moreJs-7and8.js
-    // but i used it again inside of function to show you,
-    // and in case you can run in a html seperately.
-    let products = [];
-    for(i=0; i<vehicles.length; i++){
-        products.push(' ' + vehicles[i]+'s');
-    }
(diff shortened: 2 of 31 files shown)


Week1/numberRounding.js
```diff
@@ -0,0 +1,15 @@
+//round and compair numbers
+let z = 7.25;
+console.log(z);
+let a = Math.round(z);
+console.log(a);
+let numbers = [a, z];
+let highNumber=(Math.max(z,a));
```
marcianoviereck, line 7: nice!

Week3/moreJs-10.js
```diff
@@ -0,0 +1,26 @@
...
+    // variable 'vehicles' is can be read from  moreJs-7and8.js
+    // but i used it again inside of function to show you,
+    // and in case you can run in a html seperately.
+    let products = [];
+    for(i=0; i<vehicles.length; i++){
+        products.push(' ' + vehicles[i]+'s');
+    }
+    let lastTwoProduct = products[products.length - 2] + ' and' + products[products.length - 1] ;
+    console.log(lastTwoProduct);
+    products.splice((products.length - 2), 2,);
```
marcianoviereck: will this compile? with the last `,`?

Week3/moreJs-6.js
```diff
@@ -0,0 +1,19 @@
+/*
+Create a function called vehicle, like before,
+ but takes another parameter called age, 
+ so that vehicle("blue", 1, 5) 
+ prints 'a blue used car'
+*/
+
+function vehicle (color, type, age){
+    let types = [' car' , ' motorbike'];
+    console.log('a ' + color +' ' + condition(age) + types[(type-1)]);
```
marcianoviereck: What happens if you pass '3' as value for the `type` as arguments? Will this still work?

Week3/moreJs-6.js
```diff
@@ -0,0 +1,19 @@
...
+}
+function condition(years){
```
marcianoviereck: Maybe the name `age` here is more consistent, because you use it aswell above

Week3/moreJs-9.js
```diff
@@ -0,0 +1,20 @@
...
+*/
+
+function vehicle (color, type, age){
+    let vehicles = ['motorbike' , 'caravan', 'bike', 'car'];
+    // variable 'vehicles' is can be read from  moreJs-7and8.js
+    // but i used it again inside of function to show you,
+    // and in case you can run in a html seperately.
+    console.log('a ' + color +' ' + condition(age) + vehicles[(type-1)]);
+}
+function condition(years){
```
marcianoviereck: Maybe the name age here is more consistent, because you use it aswell above

Week3/moreJs-9.js
```diff
@@ -0,0 +1,20 @@
+/*
+Change the function vehicle to use the list of question 7.
+ So that vehicle("green", 3, 1) prints "a green new bike".
+*/
+
+function vehicle (color, type, age){
+    let vehicles = ['motorbike' , 'caravan', 'bike', 'car'];
```
marcianoviereck: What happens if you pass '10' as value for the type as arguments? Will this still work?

marcianoviereck (changes requested):
Overall the homework looks good! But there are just some remarks here and there, if you can fix those then the homework is approved!:)

Week1/numberRounding.js
heartwhite: thanks.

Week3/moreJs-10.js
heartwhite: oh ! there is an extra ( unnecessary )  ' , '. 
I removed that. 
I don't know is it an error because it worked in console logs, but clearly it is weird.

Week3/moreJs-6.js
heartwhite: yes. you re right. 
I was trying to find some other parameters name. I thought it might confuse ('age' in 2 places).
but I forgot about scopes, and parameters I think.
I changed it to age. its better

Week3/moreJs-9.js
heartwhite: yes. you re right. 
I was trying to find some other parameters name. I thought it might confuse ('age' in 2 places).
but I forgot about scopes, and parameters I think.
I changed it to age. its better

Week3/moreJs-9.js
heartwhite: it returns like ' undefined ' .
I added an if statement that logs a warning when the value is out of range ( length of array).

Week3/moreJs-10.js
```diff
@@ -0,0 +1,23 @@
...
+  let products = [];
+  for (i = 0; i < vehicles.length; i++) {
+    products.push(' ' + vehicles[i] + 's');
+  }
+  let lastTwoProduct = products[products.length - 2] + ' and' + products[products.length - 1];
+  console.log(lastTwoProduct);
+  products.splice(products.length - 2, 2);
+
+  console.log("Amazing Abdullah's Garage, we sell " + products.join() + ',' + lastTwoProduct + '.');
+}
```
marcianoviereck, line 22: This is fine!

Another way to do this is this:

```

vehicles = ["car", "bike", "train"];

function advertisement() {
  let fullString = "Amazing Joe's Garage, we service ";
  for (let i = 0; i < vehicles.length; i++) {
    const element = vehicles[i];
    if (i < vehicles.length - 2) {
      fullString += vehicles[i] + ', ';
    } else if (i === vehicles.length - 2) {
      fullString += vehicles[i];
    } else {
      fullString += ' and ' + vehicles[vehicles.length - 1] + "."
    }
  }
  console.log(fullString);
}
```

marcianoviereck (approved):
Nice! The homework is approved.

Week3/moreJs-10.js
heartwhite: I think your example automate better. Thanks. it will be in my codes today before evening

PR #122, opened 2019-05-15
RiamAlali-Week3js

RiamAlali:
RiamAlali-Week3js


Week3/More-JS-16.js
```diff
@@ -0,0 +1,15 @@
...
+let o1 = { foo: "bar" };
+let o2 = { foo: "bar" };
+let o3 = o2;
+
+Show that changing o2 changes o3 (or not) and changing o1 changes o3(or not).
+Does the order that you assign (o3 = o2 or o2 = o3) matter?
+*/
+let o1 = { foo: "bar" };
+let o2 = { foo: "bar" };
+let o3 = o2;
```
marcianoviereck, line 14: This assignent is not complete yet:)

Week3/More-JS-3.js
```diff
@@ -0,0 +1,22 @@
...
+        week3 : "git",
+    }
+]
+function course(materials){
+    for(let i=0; i<materials.length;i++){
+        
+    }
+} 
+let i;
+console.log(materials);
```
marcianoviereck, line 21: Should this not be placed in line 17? (Because that line is empty)

Week3/More-JS-4.js
```diff
@@ -0,0 +1,15 @@
+
+//4 Create a function named vehicleType that receives a color, and a code, 1 for car, 2 for motorbike. And prints 'a blue motorbike' for example when called as vehicleType("blue", 2)
+
+function vehicleType(color,code){
+    if (code === 1) 
+        console.log("a " + color + " car")
+    
+    else if
+        (code===2)
+            console.log("a " + color + " motorbike");
```
marcianoviereck, line 10: nicenice, I would advise to always add brackets around your if statements, because otherwise it is easy to make silly mistakes.


```
 if (code === 1) {
        console.log("a " + color + " car")
} else if (code===2) {
        console.log("a " + color + " motorbike");
}
```

Week3/More-JS-7-8-9-10-11.js
```diff
@@ -0,0 +1,51 @@
...
+
+//8 How do you get the third element from that list?
+console.log(listOfVehicles[2]);
+
+
+
+
+/*9 Change the function vehicle to use the list of question 7.
+ So that vehicle("green", 3, 1) prints "a green new bike".*/
+function Vehicles(color,code,age){
```
marcianoviereck, line 13: This is almost correct, because right now I cannot print "a green new car". Can you change the code to allow that also (and also "a red new motorbike")

Week3/More-JS-7-8-9-10-11.js
```diff
@@ -0,0 +1,51 @@
...
+//11 What if you add one more vehicle to the list, can you have that added to the advertisement without changing the code for question 10?
+
+listOfVehicles = ["motorbike","caravan","bike","car"];
+
+function advertisement(listOfVehicles){
+    for(let i=0; i<listOfVehicles.length; i++){
+        console.log();  
+    }
+} 
+advertisement(listOfVehicles); // stucked
```
marcianoviereck, line 41: Hmm, you are on the right track!

What if you start by doing this:

`console.log(listOfVehicles[i]);`

What happens?

And what if you check if you are on the last `index` of `listOfVehicles`. If you are, then you `console.log` something, otherwise you `console.log` something else.

Maybe it's easier to help through Slack, If you are still stuck then send me a message (or any of the other teachers) I'll try to help.

Week3/strings-arrays.js
```diff
@@ -0,0 +1,37 @@
+//1.Strings
+let myString = "hello,this,is,a,difficult,to,read,sentence";
+console.log(myString);
+myString=myString.replace(/,/g, " "); // that's totally google answer i couldn't find what the g stands for
```
marcianoviereck, line 4: Ah, what the `replace` does is, replace the first `,` in your string with a ` `. And when you add the `/g` it will replace every `,` to ` `. So it will give the right result. 

Did this help? Or is it still confusing?

Week3/strings-arrays.js
```diff
@@ -0,0 +1,37 @@
...
+console.log(myString);
+
+
+
+//2 Arrays
+let favoriteAnimals = ["blowfish", "capricorn", "giraffe"];
+// add Mauro's favorite animal 'turtle' 
+favoriteAnimals[3]="turtle";
```
marcianoviereck, line 12: Ah yes, technically this will work. but the correct way to do this is like this:

```
favoriteAnimals.push('turtle');
```

See [here](https://developer.mozilla.org/nl/docs/Web/JavaScript/Reference/Global_Objects/Array/push) for more information!

task3.js
```diff
@@ -0,0 +1,19 @@
...
+//3.1 First, Declare your variable x (do not initialize it yet)
+//3.2 add a console.log statement that explains in words what you think the value of x is.
+//3.3 add a console.log statement that logs the value of x.
+//3.4 Now initialize your variable x with an integer.
+//3.5 Next, add console.log statement that explains what you think the value of x is.
+//3.6 add a console log staement that logs the value of x.
+
+let x ;
+
+console.log ("the value of my variable x will be:an Integer");
```
marcianoviereck, line 11: will the value of `x` be an integer, or will it be `undefined`?

task5.js
```diff
@@ -0,0 +1,27 @@
...
+
+console.log(a);
+
+let highest;
+if (z > a) {
+    highest = z ;
+}
+else {
+    highest = a ;
+};
```
marcianoviereck, line 23: Ah good! You can also use [math.max](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max), but this is just a tip!

task6.js
```diff
@@ -0,0 +1,16 @@
...
+
+console.log("the value of my array is : Empty.");
+
+console.log(favoriteAnimals);
+
+let myFavoriteAnimals = [ 'Deer' , 'Horse' , 'Koala' ];
+
+console.log (myFavoriteAnimals);
+
+let newMyFavoriteAnimals = myFavoriteAnimals.concat(['baby pig']);
```
marcianoviereck, line 12: Here again, you can also use `myFavoriteAnimals.push('baby pig')`

week3js-hw/More-JS-16.js
```diff
@@ -0,0 +1,15 @@
...
+let o1 = { foo: "bar" };
+let o2 = { foo: "bar" };
+let o3 = o2;
+
+Show that changing o2 changes o3 (or not) and changing o1 changes o3(or not).
+Does the order that you assign (o3 = o2 or o2 = o3) matter?
+*/
+let o1 = { foo: "bar" };
+let o2 = { foo: "bar" };
+let o3 = o2;
```
marcianoviereck, line 14: Ah did you upload all your homework twice?

marcianoviereck (changes requested):
I think you uploaded all your homework twice, overall it looks good! So keep going. There are some remarks here and there and some parts of the homework is unfinished, if you can fix the remarks (except for the tips) then the homework will be approved:). If you need help just send a message of Slack or so

PR #125, opened 2019-05-15
7alip - Homework Week 3

7alip:


7alip_week2/strings-arrays/app.js
```diff
@@ -0,0 +1,64 @@
...
+  If you assign new pushed array to a new variable, it returns a number. if you want to log desired result, you need to call it without assigning. 
+  */
+
+  let jimsAnimal = 'meerkat';
+
+  // Here is the solution https://stackoverflow.com/a/12710609
+  Array.prototype.insert = function (index, item) {
+    this.splice(index, 0, item);
+  }
+
```
PKnar: This is a good solution, however, what this solution does is, it copies the original array and then edits that copied version, leaving the original array unchanged. For this homework you could just use splice(), it is faster and it directly changes the original array, which I think is what homework requires.

7alip_week2/more-js/app.js
```diff
@@ -0,0 +1,207 @@
...
+function showObj(obj) {
+  let objKeys = Object.keys(obj);
+
+  let objVals = Object.values(obj);
+
+  for (let i = 0; i < objKeys.length; i++) {
+    console.log(`The value of the property '${objKeys[i]}' is '${objVals[i]}'`)
+  }
+}
+
```
PKnar, line 25: This works, but check out `for...in` loop for a shorter solution.

7alip_week2/more-js/app.js
```diff
@@ -0,0 +1,207 @@
...
+      str += `${currentEl}s, `;
+    } else {
+      str += `${lastEl}s.`;
+    }
+  }
+
+  return str;
+}
+
+console.log(ad(adStr, vehicles)); // Amazing Joe's Garage, we service motorbike, caravan, car, truck.
```
PKnar, line 110: According to the instruction it should be ` Amazing Joe's Garage, we service motorbike, caravan, car and truck.`, So you just need to add `and` before the last word. Overall good job!

PKnar (commented):
Well done :+1:   The pull request looked a bit messy, I hope I did not miss anything. There were one-two solutions that could have been done in a shorter way but overall you did a good job, especially in building reusable functions. Keep up the good work

PR #127, opened 2019-05-15
hani's homework  java script 1 week3

hanialkudsi:


new.js
```diff
@@ -0,0 +1,161 @@
...
+    if(type == 1 && age == 1){
+        console.log("a new " + color + " car");
+    }else if(type == 1 && age == 2){
+        console.log("a used " + color + " car");
+    }else if (type == 2 && age == 1){
+        console.log("a new " + color + " motorbike");
+    }else if (type == 2 && age == 2){
+        console.log("a used " + color + " motorbike");
+    }
+}
```
PKnar, line 99: Here you repeat yourself, think about how to make this code better. What if the car is 3,4,5,6... years old? You learned about single line if statement, see if you can do something with it.

new.js
```diff
@@ -0,0 +1,161 @@
...
+        console.log("a used " + color + " " + listOfVehicles[0]);
+    }else if (type == 2 && age == 1){
+        console.log("a new " + color + " " + listOfVehicles[1]);
+    }else if (type == 2 && age == 2){
+        console.log("a used " + color + " " + listOfVehicles[1]);
+    }
+}
+vehicle("blue",1,1);
+vehicle("green",2,2);
+
```
PKnar, line 129: Read my comment above and try to implement the same idea here

new.js
```diff
@@ -0,0 +1,161 @@
...
+        adv +=  " and " + listOfVehicles[i] + ".";
+      }else if(i == listOfVehicles.length-2){
+         adv += listOfVehicles[i];
+      }
+      else { adv += listOfVehicles[i] +  ", " ;
+      }
+}console.log(adv)
+}
+advertisement();
+
```
PKnar, line 148: Nice! but you forgot to make each vehicle plural :smiley:

PKnar (commented):
Overall you did pretty well! I see you committed 1 file only, it seems like the rest of the homework is missing?

PR #128, opened 2019-05-18
JS1/week3

Shvan1:


week3.js
```diff
@@ -0,0 +1,131 @@
...
+let favoriteAnimalsOfJim = ["blowfish", "meerkat", "capricorn", "giraffe", "turtle"];
+console.log("The favorite animal of Jim is added to the array.");
+console.log(favoriteAnimalsOfJim);
+console.log("The array has a length of:", favoriteAnimalsOfJim.length);
+
+let favoriteAnimalsOfJason = ["blowfish", "meerkat", "capricorn", "turtle"];
+console.log(favoriteAnimalsOfJason);
+console.log(favoriteAnimalsOfJason.length);
+console.log(favoriteAnimalsOfJason[1], "The item you are looking for is at index:2");
+
```
PKnar, line 23: This 'Arrays' section needs to be changed. I guess you did not understand the instructions correctly. For this section, the idea was to find array methods that allow you to add/remove/insert items to the array, meaning you should not write those arrays yourself but let the array methods modify it for you.  Start, for example by learning push() method https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push

week3.js
```diff
@@ -0,0 +1,131 @@
...
+function colorCar(color){
+    return "A" + color + "car";
+ }
+ console.log(colorCar("red"));
+//3.
+let country = {Land: "The Netherlands", Population:"17 million", Capital:"Amsterdam"};
+console.log(country);
+console.log(country.Land);
+console.log(country.Population);
+//4.
```
PKnar, line 39: The homework instruction is `Create an object and a function that takes the object as a parameter and prints out all of its properties and values.`  you created the object but not the function. Here you console.log everything yourself. Build a function and let it do the hard work. Check out  `for..in` loop as a hint and try to build a function that homework requires.  Read more about for..in loop here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in

week3.js
```diff
@@ -0,0 +1,131 @@
...
+//9
+function newVehicle(color){
+    if (vehicleList[2] = 3){
+        console.log("A" + "" + color + "" + "new" + "" + "bike");
+    } else (vehicleList[2] > 3);{
+        console.log("A" + "" + color + "" + "old" + "" + "bike")
+    }
+}
+newVehicle("green", 3, 1);
+
```
PKnar, line 78: The  `newVehicle`  function needs to have three parameters, and you need to use those parameters to make your function reusable. Now your function uses 'bike' as the only vehicle type, what if we want to print  'an old black caravan ' or  ' a new red car'?

week3.js
```diff
@@ -0,0 +1,131 @@
...
+if (advertisement.length = 5) {
+    console.log("Amazing Joe's Garage >>> Here we service cars, motorbikes, caravans and bikes.");
+}else {
+    console.log("Amazing Joe's Garage >>> Here we service bulldozers, lorries and trucks.");
+}
+//11
+const newVehicleList = advertisement + "" + "scooter";
+console.log(newVehicleList);
+//12.
+let object;
```
PKnar: here you declared a variable that has no type.  if you console.log it, the output will be 'undefined', but the instruction says `declare an empty object`

week3.js
```diff
@@ -0,0 +1,131 @@
...
+//13+14
+let teachers = {Philip: "HTML/CSS", Rob: "HTML/CSS", Hardit: "CLI", Bonan: "CLI", Ogor: "CLI", Unmesh: "GIT", Marciano: "JS", Sander: "JS"};
```
PKnar: At first, you need to create an object that contains teachers' name only. Then you should add the languages they teach, separately. For example,  let's say you have  `let person ={}` object, if you want to add a new property you can do  person.name = 'Anna',  or person.skills=[HTML,CSS,JS]. Use this idea and follow the instructions to do this part correctly

week3.js
```diff
@@ -0,0 +1,131 @@
...
+    console.log(advertisement[index]);
+}
+if (advertisement.length = 5) {
+    console.log("Amazing Joe's Garage >>> Here we service cars, motorbikes, caravans and bikes.");
+}else {
+    console.log("Amazing Joe's Garage >>> Here we service bulldozers, lorries and trucks.");
+}
+//11
+const newVehicleList = advertisement + "" + "scooter";
+console.log(newVehicleList);
```
PKnar: I would suggest you read about `Don't Repeat Yourself (DRY) coding` and `hard coding`. Your solution is an example of `hard coding`,  read about it and see if you can solve this in a better way.Good luck

PKnar:
I see you misunderstood some of the homework instructions which resulted in a lot of  `hard coding`, this means you often repeated yourself and did not create reusable functions that would do the hard work for you. Like I suggested in one of my comments, it's a nice idea to read about hard-coding and DRY-coding and if possible, try to do the homework again. Good luck!

PKnar (changes requested):
I see you misunderstood some of the homework instructions which resulted in a lot of  `hard coding`, this means you often repeated yourself and did not create reusable functions that would do the hard work for you. Like I suggested in one of my comments, it's a nice idea to read about hard-coding and DRY-coding and if possible, try to do the homework again. Good luck!

PKnar (changes requested):
I see you misunderstood some of the homework instructions which resulted in a lot of  `hard coding`, this means you often repeated yourself and did not create reusable functions that would do the hard work for you. Like I suggested in one of my comments, it's a nice idea to read about hard-coding and DRY-coding and if possible, try to do the homework again. Good luck!

week3.js
Shvan1: Thanks for reviewing, gonna check it again

Shvan1: Changed to: let favoriteAnimals = ["blowfish", "capricorn", "giraffe"];
console.log(favoriteAnimals.push("turtle"));
console.log(favoriteAnimals);

let addedAnimalOfJim = ["blowfish", "capricorn", "giraffe", "turtle"];
console.log(addedAnimalOfJim.splice(1, 0, "meerkat"));
console.log("The new value will be: ["blowfish", "meerkat", "capricorn", "giraffe", "turtle"]");
console.log(addedAnimalOfJim);
console.log("The array has a length of:", addedAnimalOfJim.length);

let deletedAnimalOfJason = ["blowfish", "meerkat", "capricorn", "giraffe", "turtle"];
console.log(deletedAnimalOfJason.splice(-2, 1,));
console.log(deletedAnimalOfJason.length);
console.log(deletedAnimalOfJason[1], "The item you are looking for is at index:2");

week3.js
Shvan1: Changed to: let country = {Land: "The Netherlands", Population:"17 million", Capital:"Amsterdam"};
for (const elements in country) {
    console.log(country.elements = country[elements]);
}

week3.js
Shvan1: changed to: let empty = {};

week3.js
Shvan1: Changed to: const newVehicle = advertisement;
console.log(newVehicle.push("scooter"));
consol.log(newVehicle);

week3.js
Shvan1: Changed to: let teachers = {
    name : ["Rob", "Philip", "Unmesh"]
    };
console.log(teachers);
let task = teachers.skills = ["HTML/CSS", "HTML/CSS", "JavaScript"];
console.log(task);

week3.js
Shvan1: Changed to: function newVehicle(color, code, age){
    if (vehicleList[2] <= 3){
        console.log("A" + " " + color + " " + "new" + " " + "bike");
    } else (vehicleList[2] > 3);{
        console.log("A" + " " + color + " " + "old" + " " + "bike")
    }
}
newVehicle("green", 3, 1);

week3.js
PKnar: Check out this array method, see how you can use it in your code https://developer.mozilla.org/nl/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf

week3.js
PKnar: You are almost there, the output should be  Land = The Netherlands
Population = 17 million
Capital = Amsterdam

week3.js
PKnar: You have not used code and age parameters. This function still needs to be improved. You still use 'bike' as the only vehicle type

PR #132, opened 2019-07-30
adding homework files

ersarue:
homework 1 to 7 of week2,

### DIFF
--- a/Week2/1.js
+++ b/Week2/1.js
@@ -1,5 +1,5 @@
-let dutch = 'hoi';
-let turkish = 'merhaba';
-let spanish = 'hola';
+const dutch = 'hoi';
+const turkish = 'merhaba';
+const spanish = 'hola';
 
 console.log('saying hello in different languages: ' + dutch, turkish, spanish);
--- a/Week2/1.js
+++ b/Week2/1.js
@@ -1,0 +1,5 @@
+let dutch = 'hoi';
+let turkish = 'merhaba';
+let spanish = 'hola';
+
+console.log('saying hello in different languages: ' + dutch, turkish, spanish);
--- a/Week2/3.js
+++ b/Week2/3.js
@@ -1,0 +1,11 @@
+let x;
+
+console.log('the value of variable x will be a number');
+
+console.log(x);
+
+x = 4;
+
+console.log('the value of variable x will be number 4');
+
+console.log(x);
--- a/Week2/3.js
+++ b/Week2/3.js
@@ -1,8 +1,10 @@
 let x;
 
-console.log('the value of variable x will be a number');
+console.log('the value of variable x will be an undefined');
 
 console.log(x);
+
+// returned: undefined
 
 x = 4;
 
--- a/Week2/5.js
+++ b/Week2/5.js
@@ -1,0 +1,17 @@
+let z = 7.25;
+
+console.log(z);
+
+let a = Math.round(z);
+
+console.log(a);
+
+// to compare and see the values :
+
+console.log(a > z);
+
+// returned false. so a is not greater than z.
+
+let highestNumber = z;
+
+console.log(highestNumber);
--- a/Week2/5.js
+++ b/Week2/5.js
@@ -2,16 +2,14 @@ let z = 7.25;
 
 console.log(z);
 
+// returned: 7.25
+
 let a = Math.round(z);
 
 console.log(a);
 
-// to compare and see the values :
+// returned: 7
 
-console.log(a > z);
-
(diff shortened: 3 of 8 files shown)


Week2/1.js
```diff
@@ -0,0 +1,5 @@
+let dutch = 'hoi';
+let turkish = 'merhaba';
+let spanish = 'hola';
+
+console.log('saying hello in different languages: ' + dutch, turkish, spanish);
```
marcianoviereck, line 5: Do you think it's better to use `let` here, or `const`?

Week2/3.js
```diff
@@ -0,0 +1,11 @@
+let x;
+
+console.log('the value of variable x will be a number');
```
marcianoviereck: are you sure that the value is a number? I don't think it will be

Week2/5.js
```diff
@@ -0,0 +1,17 @@
...
+
+console.log(z);
+
+let a = Math.round(z);
+
+console.log(a);
+
+// to compare and see the values :
+
+console.log(a > z);
```
marcianoviereck: The assignment is as follows:

```
5.5 So now we have z and a find a way to compare the two values and store the highest of the two in a new variable.
5.6 console.log the highest value.
```

So can you store the highest value in a variable and then console.log that one?

Week2/6.js
```diff
@@ -0,0 +1,11 @@
+myArray = [];
```
marcianoviereck: are you sure that this will work? (without `const` or `let`)?

Week2/6.js
```diff
@@ -0,0 +1,11 @@
...
+
+console.log(myArray);
+
+myFavoriteAnimals = ['cats', 'rabbits', 'horses'];
+
+console.log(myFavoriteAnimals);
+
+myFavoriteAnimals.push('baby pig');
```
marcianoviereck, line 9: nice solution!

marcianoviereck (commented):
Looks good! I do have some remarks here and there:) But overall it's really nice!

marcianoviereck (changes requested):
forgot to mark this PR as 'request changes'

Week2/1.js
ersarue: it is better to use `const` in this case

Week2/3.js
ersarue: I misunderstood it, you're right. it should have been 'the value of variable x will be an integer'.

Week2/5.js
ersarue: yes, it is explicit after doing console.log(a > z) that z has bigger value than a. and z is assigned to `highestNumber` in order to store highest value in a new variable. do I get it wrong Marciano?

Week2/6.js
ersarue: it is completely overlooked issue in a busy time. so it should be `let myArray = [];` or `const myArray = [];`.

Week2/5.js
ersarue: I just figured out that `Math.max()`  can be used to store the highest value of variables. so, `let highestNumber = Math.max(z ,a)` would give us what is wanted.

Week2/1.js
marcianoviereck: Yeah true, can you change it in the code?

Week2/3.js
marcianoviereck: Ah, close! you can `console.log` it for yourself and see the value:)

Week2/5.js
marcianoviereck: Ah, yeah your code is techically correct. But you are deciding for yourself which value is higher.
You do this

```
console.log(a > z);
```

But, you don't store the `highest number` in a variable.

And after that you do this 
`let highestNumber = z;`

But how does your code know that `z` is higher than `a`? Do you understand what I mean?

And yeah, if you do this:
`let highestNumber = Math.max(z ,a)` then the code knows what the highest value is, because it is stored in `highestNumber`. So the differnce is that now the code decides that the value is higher, but before, you decided which value is higher.

If it's not clear I will try to explain it better:)

Week2/6.js
marcianoviereck: I think `const` is fine

marcianoviereck (changes requested):
Can you update your code with the changes? Then I will approve it:)

ersarue:
thanks for the feedbacks! I'll make the required changes and a PR soon.

ersarue (commented):
made the corrections, please have a one more look

marcianoviereck (approved):
Great! The homework is approved!

PR #133, opened 2019-07-30
my JS week2 home work

husenjanobul:
this is js1 week2 home work .
the week1 home work was ,  we have done in the class.

### DIFF
--- a/Week2/arrays.js
+++ b/Week2/arrays.js
@@ -1,4 +1,5 @@
 var myArrys = [];
+console.log("i think the value of myArrys is : length 0 , []  .");
 console.log("arrays are used to store multiple values in a single variable.");
 console.log(myArrys);
 
--- a/Week2/arrays.js
+++ b/Week2/arrays.js
@@ -1,9 +1,9 @@
-var myArrys = [];
+let myArrys = [];
 console.log("i think the value of myArrys is : length 0 , []  .");
 console.log("arrays are used to store multiple values in a single variable.");
 console.log(myArrys);
 
-var petsWorld = ["cat", "dog", "chicken", "fish" ];
+let petsWorld = ["cat", "dog", "chicken", "fish" ];
 console.log(petsWorld);
 petsWorld.push("horse");
 console.log(petsWorld);
--- a/Week2/arrays.js
+++ b/Week2/arrays.js
@@ -1,0 +1,8 @@
+var myArrys = [];
+console.log("arrays are used to store multiple values in a single variable.");
+console.log(myArrys);
+
+var petsWorld = ["cat", "dog", "chicken", "fish" ];
+console.log(petsWorld);
+petsWorld.push("horse");
+console.log(petsWorld);
--- a/Week2/assign-string.js
+++ b/Week2/assign-string.js
@@ -1,0 +1,7 @@
+console.log("string is zero or more characters written inside quotes.");
+var y = "huseyin";
+console.log(y);
+
+y = "obul";
+console.log("the value of my string will be : obul");
+console.log(y);
--- a/Week2/assign-string.js
+++ b/Week2/assign-string.js
@@ -1,5 +1,6 @@
 console.log("string is zero or more characters written inside quotes.");
-var y = "huseyin";
+let y = "huseyin";
+console.log("the value of my string will be : huseyin");
 console.log(y);
 
 y = "obul";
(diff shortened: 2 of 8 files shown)


Week2/arrays.js
```diff
@@ -0,0 +1,8 @@
+var myArrys = [];
+console.log("arrays are used to store multiple values in a single variable.");
```
marcianoviereck, line 3: yeah that is correct, but what do you think the value of the `myArrs`?

Week2/arrays.js
```diff
@@ -0,0 +1,8 @@
+var myArrys = [];
```
marcianoviereck: It's better to use `let` or `const` instead of `var`, so can you change that:)?

Week2/assign-string.js
```diff
@@ -0,0 +1,7 @@
+console.log("string is zero or more characters written inside quotes.");
+var y = "huseyin";
+console.log(y);
```
marcianoviereck, line 4: Ah, here again, your explanation is correct, but waht do you think the value of `y` is?

Week2/assign-string.js
```diff
@@ -0,0 +1,7 @@
+console.log("string is zero or more characters written inside quotes.");
+var y = "huseyin";
```
marcianoviereck: Again the same comment with the `var` :)

Week2/round-number.js
```diff
@@ -0,0 +1,18 @@
...
+console.log(z);
+var a = 7;
+console.log(a);
+console.log("the highest value is : 7.25");
+
+// second
+var a = Math.round(z);
+console.log(a);
+var c = z - a;
+if ( c > 0 && z != a ){
```
marcianoviereck: Ah wow, cool solution!
There is a much easier way to do this, you can use [Math.max](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max) it makes this a lot easier

marcianoviereck (commented):
The homework looks really nice and some solutions are pretty cool! But can you change the `var`'s and look at my other comments?:)

marcianoviereck (changes requested):
Forgot to mark this PR as 'request changes'

Week2/arrays.js
husenjanobul: now i made it .

Week2/arrays.js
husenjanobul: thank you, i changed var to let.

Week2/round-number.js
husenjanobul: i will try this.

husenjanobul: @marcianoviereck  thanks for feedback.

Week2/round-number.js
```diff
@@ -0,0 +1,19 @@
...
```
marcianoviereck: Ah your new solution (on line 18) is fine, but can you remove this code then?

marcianoviereck (changes requested):
The homework is almost approved, can you only remove the old code in the `round-number.js`?

Week2/declare-integer.js
```diff
@@ -0,0 +1,11 @@
+var x;
```
marcianoviereck: Can you change this `var` to `let` or `const`?

Week2/round-number.js
```diff
@@ -0,0 +1,19 @@
+var z = 7.25;
```
marcianoviereck: Can you change this `var` to `let` or `const`?

Week2/round-number.js
```diff
@@ -0,0 +1,19 @@
...
+console.log(z);
+var a = 7;
```
marcianoviereck: Can you change this `var` to `let` or `const`?

Week2/round-number.js
```diff
@@ -0,0 +1,19 @@
...
+console.log(a);
+console.log("the highest value is : 7.25");
+
+// second
+var a = Math.round(z);
```
marcianoviereck: Can you change this `var` to `let` or `const`?

marcianoviereck (changes requested):
And can you also cahnge your other `var`s to `let` or `const`?

marcianoviereck (approved):
Ah nice! The homework is approved now!

PR #134, opened 2019-07-31
adding hw of week2

FarukTercan:
Hi, this is my Homework for Javascript1 week2.

### DIFF
--- a/Week2/1.Hello World.js
+++ b/Week2/1.Hello World.js
@@ -1,0 +1,5 @@
+// 1
+let language1 = 'Merhaba dunya!' ,
+    language2 = 'Hello world!' ,
+    language3 = 'Hola, mundo!' ;
+console.log(language1, language2, language3) ;
--- a/Week2/3.Declare-Initialize Variable.js
+++ b/Week2/3.Declare-Initialize Variable.js
@@ -1,0 +1,17 @@
+// 3.1
+let x ;
+
+// 3.3
+console.log("the value of my variable x will be: 7 ");
+
+// 3.3
+console.log(x) ;
+
+// 3.4
+x=7;
+
+// 3.5
+console.log("the value of my variable x:7");
+
+// 3.6
+console.log(x) ;
--- a/Week2/5.Round-Compare.js
+++ b/Week2/5.Round-Compare.js
@@ -1,0 +1,25 @@
+// 5.1
+let z = 7.25 ;
+
+// 5.2
+console.log(z) ;
+
+// 5.3
+let a = Math.round(z) ;
+
+// 5.4
+console.log(a) ;
+
+// 5.5
+if (a > z) {
+    theHighestOne = a ;
+}
+    else {
+    theHighestOne = z ;    
+}
+
+// 5.6
+console.log(theHighestOne) ;
+let bestTravelCameras = [] ;
+console.log("the value of bestTravelCameras :") ;
+console.log(bestTravelCameras) ;
--- a/Week2/6.Arrays.js
+++ b/Week2/6.Arrays.js
@@ -1,0 +1,23 @@
+// 6.1
+let animals = [] ;
+
+// 6.2
+console.log("the value of my array : Samoyed dog, British Shorthair, Horse") ;
+
+// 6.3
+console.log(animals) ;
+
+// 6.4
+let myFavoriteAnimals = ['Samoyed dog' , 'British Shorthair' , 'Horse'] ;
+
+// 6.5
+console.log(myFavoriteAnimals) ;
+
+// 6.6
+myFavoriteAnimals.push("Daan's favorite animal ('baby pig')") ;
+//  or we can add it as a new variable 
+// let daansFavoriteAnimal = ['baby pig'] ;
+// myFavoriteAnimals.push(daansFavoriteAnimal[0]);
+
+// 6.7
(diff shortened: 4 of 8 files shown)


Week2/1.Hello World.js
```diff
@@ -0,0 +1,5 @@
+// 1
+let language1 = 'Merhaba dunya!' ,
```
marcianoviereck, line 2: Can you also add which languages these are?

And do you think it's better to use `let` or `const`?

Week2/3.Declare-Initialize Variable.js
```diff
@@ -0,0 +1,17 @@
+// 3.1
+let x ;
+
+// 3.3
+console.log("the value of my variable x will be: 7 ");
```
marcianoviereck, line 5: are you sure the value is 7? (I know it will be 7, but right now I't not 7 yet)

Week2/5.Round-Compare.js
```diff
@@ -0,0 +1,25 @@
...
+console.log(z) ;
+
+// 5.3
+let a = Math.round(z) ;
+
+// 5.4
+console.log(a) ;
+
+// 5.5
+if (a > z) {
```
marcianoviereck, line 14: Ah nicenice! You know that you can also use [Math.max](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max) (it makes this a lot easier!)

Week2/5.Round-Compare.js
```diff
@@ -0,0 +1,25 @@
...
+    theHighestOne = a ;
+}
+    else {
+    theHighestOne = z ;    
+}
+
+// 5.6
+console.log(theHighestOne) ;
+let bestTravelCameras = [] ;
```
marcianoviereck: I don't think this is part of `5` right?

Week2/6.Arrays.js
```diff
@@ -0,0 +1,23 @@
+// 6.1
+let animals = [] ;
+
+// 6.2
+console.log("the value of my array : Samoyed dog, British Shorthair, Horse") ;
```
marcianoviereck, line 5: Is the value of the array that value right now? What is the value of `animals` right now? if I `console.log` it?

Week2/6.Arrays.js
```diff
@@ -0,0 +1,23 @@
...
+let myFavoriteAnimals = ['Samoyed dog' , 'British Shorthair' , 'Horse'] ;
+
+// 6.5
+console.log(myFavoriteAnimals) ;
+
+// 6.6
+myFavoriteAnimals.push("Daan's favorite animal ('baby pig')") ;
+//  or we can add it as a new variable 
+// let daansFavoriteAnimal = ['baby pig'] ;
+// myFavoriteAnimals.push(daansFavoriteAnimal[0]);
```
marcianoviereck, line 20: Nice! You can also just do `myFavoriteAnimals.push('baby pig')` that's a bit easier!

marcianoviereck (changes requested):
The homework looks nice, I do have some remarks here and there!

Week2/1.Hello World.js
```diff
@@ -0,0 +1,5 @@
...
+    language2 = 'Hello world!' ,
+    language3 = 'Hola, mundo!' ;
+console.log(language1, language2, language3) ;
```
marcianoviereck: nice, but can you also add which languages these are:)?

Week2/3.Declare-Initialize Variable.js
marcianoviereck: Are you sure that the value is 7?

Week2/1.Hello World.js
marcianoviereck: And do you think it's better to use `let ` or `const`?

Week2/5.Round-Compare.js
```diff
@@ -0,0 +1,25 @@
...
+// 5.4
+console.log(a) ;
+
+// 5.5
+if (a > z) {
+    theHighestOne = a ;
+}
+    else {
+    theHighestOne = z ;    
+}
```
marcianoviereck: A goodgood, there is an easier solution though, with [Math.max](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max)

Week2/5.Round-Compare.js
marcianoviereck, line 23: I don't think that this is part of assignment 5, right?

Week2/6.Arrays.js
marcianoviereck: Are you sure that this is the value?

Week2/6.Arrays.js
```diff
@@ -0,0 +1,23 @@
...
+console.log(animals) ;
+
+// 6.4
+let myFavoriteAnimals = ['Samoyed dog' , 'British Shorthair' , 'Horse'] ;
+
+// 6.5
+console.log(myFavoriteAnimals) ;
+
+// 6.6
+myFavoriteAnimals.push("Daan's favorite animal ('baby pig')") ;
```
marcianoviereck: You can also do it easier, with `myFavoriteAnimals.push('baby pig')` ;)

marcianoviereck (changes requested):
The homework looks fine! I do have some remarks though

PR #136, opened 2019-07-31
Homework JS1 Week 2

ihk2018:


Week2/eight.js
```diff
@@ -0,0 +1,75 @@
+// 8.1
+
+let myfavoriteanimals = ['lions', 'horses','eagles'];
```
marcianoviereck: small remark, it's better to use `camelCase`, so then it's better to write `myFavouriteAnimals` instead of `myfavouriteanimals` because it's easier to read.
The same goes for your other variables

Week2/eight.js
```diff
@@ -0,0 +1,75 @@
...
+    console.log ('Same Type');
+} else (typeof myfavoritetennisplayer != typeof  myage); {
+    console.log ('Different Type');
+}
+
+if (typeof myage == typeof  answer) { 
+    console.log ('Same Type');
+} else (typeof myage != typeof  answer);{
+    console.log ('Different Type');
+}
```
marcianoviereck, line 77: nice! looks good

Week2/five.js
```diff
@@ -0,0 +1,20 @@
...
+// 5.3
+
+let a = Math.round(z);
+
+// 5.4
+console.log(a);
+
+// 5.5 find a way to compare the two values and store the highest of the two in a new variable.
+
+let highest = Math.max(z,a);
```
marcianoviereck, line 16: clean solution!

Week2/four.js
```diff
@@ -0,0 +1,21 @@
+// 4
+let y = "Hey Guys";
+
+// 4.1
+
+console.log("the value of the string will be: undefined ");
```
marcianoviereck: Are you sure that this is what the value will be?

Week2/six.js
```diff
@@ -0,0 +1,28 @@
...
+
+animals=['horse', 'lion', 'eagle'];
+
+// 6.5
+
+console.log(animals);
+
+// 6.6 
+
+animals [3] = 'baby pig';
```
marcianoviereck: Ah, this works technically yeah, but it's cleaner/easier to use [push](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push)

marcianoviereck (changes requested):
Homework looks good, and nice that you've already finished untill 8! I do have some remarks though:)

Week2/eight.js
ihk2018: Dear @marcianoviereck ;  just reviewed the freecodecamp basicjavascript part, and read camelCase there as weel , from now on I need to  pay attention to that.  So;
myfavoriteanimals should be myFavoriteAnimals; 
myfavoritetennisplayer should be myFavoriteTennisPlayer

thanks for the feedback

Week2/eight.js
ihk2018: thanks :)

Week2/four.js
ihk2018: if i write ; console.log (y) and run the program , then the value will be hey guys

Am I right?

Week2/eight.js
marcianoviereck: Ah that's good!

Week2/four.js
marcianoviereck: You are right, so you can update your code (or change the "hey guys")

Week2/six.js
ihk2018: i did not know JavaScript Array push() Method; just reviewed it ; so it would be better if i write as below;

animals=['horse', 'lion', 'eagle'];
 ["horse", "lion", "eagle"]
animals.push('baby pig');
console.log(animals);
["horse", "lion", "eagle", "baby pig"]

marcianoviereck: Yeah that looks good! Can you update your code then I can review it again?

marcianoviereck (approved):
Nice! The homework is approved now:)

PR #137, opened 2019-07-31
upload js files of homework js1 week2

malnaqeeb:
Please let me know if there is anything have to be changed :)

### DIFF
--- a/Week2/10_program.js
+++ b/Week2/10_program.js
@@ -1,4 +1,17 @@
-let myArray =["My age","30"]; // With array is possibale to store different data, within a single variable
+// 10.1 Can you store multiple types in an array? Numbers and strings? Make an example that illustrates your answer.
+
+let myArray =["My age",30 , ["Student" ,5]]; // With array is possibale to store different data, within a single variable
 console.log(myArray);
-let maxNumber = Math.pow(10, 1000);
-console.log(maxNumber);
+
+
+//10.2 Can you compare infinities? (Not in Eyad's world) - does 6/0 === 10/0? How can you test this?
+//10.3 Add console.log statements to the above program
+let inf1 = 10/0;
+let inf2 = 6/0;
+
+function comparInfinities(inf1,inf2){
+        if (inf2 = Number.MAX_VALUE , inf2 = Number.MAX_VALUE)
+        console.log("inf1 and inf2 are infinity")
+        else console.log("False")
+}
+comparInfinities()
--- a/Week2/10_program.js
+++ b/Week2/10_program.js
@@ -1,17 +1,14 @@
 // 10.1 Can you store multiple types in an array? Numbers and strings? Make an example that illustrates your answer.
 
-let myArray =["My age",30 , ["Student" ,5]]; // With array is possibale to store different data, within a single variable
+let myArray = ["My age", 30, ["Student", 5]]; // With array is possibale to store different data, within a single variable
 console.log(myArray);
 
 
 //10.2 Can you compare infinities? (Not in Eyad's world) - does 6/0 === 10/0? How can you test this?
 //10.3 Add console.log statements to the above program
-let inf1 = 10/0;
-let inf2 = 6/0;
+let inf1 = 10 / 0;
+let inf2 = 6 / 0;
 
-function comparInfinities(inf1,inf2){
-        if (inf2 = Number.MAX_VALUE , inf2 = Number.MAX_VALUE)
+if (inf2 === inf2)
         console.log("inf1 and inf2 are infinity")
-        else console.log("False")
-}
-comparInfinities()
+else console.log("False")
--- a/Week2/10_program.js
+++ b/Week2/10_program.js
@@ -1,0 +1,4 @@
+let myArray =["My age","30"]; // With array is possibale to store different data, within a single variable
+console.log(myArray);
+let maxNumber = Math.pow(10, 1000);
+console.log(maxNumber);
--- a/Week2/1_hello_world.js
+++ b/Week2/1_hello_world.js
@@ -1,0 +1,8 @@
+console.log("Hallo Wereld") // Dutch
+console.log("Bonjour le monde") // French
+console.log("مرحبا بالعالم") // Arabic
+console.log("ہیلو دنیا") // Urdu
+console.log("Selam Dünya") // Turkish
+console.log("你好，世界") // Chinese
+console.log("Hallo Welt") // German
+
--- a/Week2/1_hello_world.js
+++ b/Week2/1_hello_world.js
@@ -1,8 +1,8 @@
-console.log("Hallo Wereld") // Dutch
-console.log("Bonjour le monde") // French
-console.log("مرحبا بالعالم") // Arabic
-console.log("ہیلو دنیا") // Urdu
-console.log("Selam Dünya") // Turkish
-console.log("你好，世界") // Chinese
-console.log("Hallo Welt") // German
+console.log("Hallo Wereld"); // Dutch
+console.log("Bonjour le monde"); // French
+console.log("مرحبا بالعالم"); // Arabic
+console.log("ہیلو دنیا"); // Urdu
+console.log("Selam Dünya"); // Turkish
+console.log("你好，世界"); // Chinese
+console.log("Hallo Welt"); // German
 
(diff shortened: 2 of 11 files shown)


Week2/1_hello_world.js
```diff
@@ -0,0 +1,8 @@
+console.log("Hallo Wereld") // Dutch
```
marcianoviereck: Goodgood, don't forget to add the `;` at the end (also for the rest of your code).

This might seem useless, but it will save you a lot of headache later on:)

Week2/2_SyntaxError.js
```diff
@@ -0,0 +1,8 @@
+console.log('I\'m awesome')
+ // Error type:  SyntaxError: missing ) after argument list } 
+
+ // Reason    :  here is no "+" operator to concatenate the string
+ //              so JavaScript expects the argument for the log function to be just 'I'. 
+ //              In that case, it should be terminated by a closing parenthesis.
```
marcianoviereck, line 6: Nice explanation!

Week2/5_round_numbder.js
```diff
@@ -0,0 +1,6 @@
+let z = 7.25;
```
marcianoviereck: Do you think it's better to use `let` or `const` here?

Week2/5_round_numbder.js
```diff
@@ -0,0 +1,6 @@
...
+console.log(z);
+let a = Math.round(z);
+console.log(a);
+let BiggerNum = Math.max(a,z);
+console.log(BiggerNum);
```
marcianoviereck: nice! Please use lowercase for your variablename, so `biggerNum` instead of `BiggerNum`, see [naming conventions](https://github.com/HackYourFuture/fundamentals/blob/master/fundamentals/naming_conventions.md)

Week2/6_arry.js
```diff
@@ -0,0 +1,8 @@
+let zoo=[];
+console.log(zoo);
+console.log("Animals that I saw in the Zoo");
```
marcianoviereck: Is this the value of `zoo`? I think that in the assignment you have to write down the value

Week2/6_arry.js
```diff
@@ -0,0 +1,8 @@
...
+zoo=['Lione', 'Tiger', 'Giraffe', 'Elephant'];
+console.log(zoo);
+zoo[4]=('baby pig');
```
marcianoviereck: Technically this works fine! But it's better to use [push](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push)

marcianoviereck (changes requested):
Looks good! I do have some remarks though:)

Week2/1_hello_world.js
malnaqeeb: Oky, i did

Week2/2_SyntaxError.js
malnaqeeb: Thanx  :+1:

Week2/5_round_numbder.js
malnaqeeb: Hmm, Now i think const is better because it's means that the identifier can’t be reassigned. Is that correct?

Week2/5_round_numbder.js
malnaqeeb: Done :+1:

Week2/6_arry.js
malnaqeeb: I got confused about this points.
>  According to the assignment i have to write what the value of the array is.
> But on this step (6.2) the array is still empty and if i run **console.log(zoo)** the result will be [ ]
So is this what i have to write as value of the array ?

Week2/6_arry.js
malnaqeeb: Done :+1:

malnaqeeb (commented):
Thanx a lot  @marcianoviereck, I have changed the codes according to your feedback :+1:

Week2/5_round_numbder.js
marcianoviereck: Yeah

Week2/6_arry.js
marcianoviereck: Yeah, exactly! That is the value

Week2/10_program.js
```diff
@@ -0,0 +1,4 @@
+let myArray =["My age","30"]; // With array is possibale to store different data, within a single variable
```
marcianoviereck: The assignment goes as follows:

> Can you store multiple types in an array

And you've made an array, that's fine! But you only store one type in it (`String`). Can you change it so that it will contain multiple `types` of values?

Week2/10_program.js
```diff
@@ -0,0 +1,4 @@
...
+console.log(myArray);
+let maxNumber = Math.pow(10, 1000);
```
marcianoviereck: The assignment is as follows:

>  Can you compare infinities? (Not in Eyad's world) - does 6/0 === 10/0? How can you test this?

I don't think this is the right way to `compare` infinities. Do you understand what the assignment is? Or should I try to explain it differently?

Week2/8_checkTypeOfData.js
```diff
@@ -0,0 +1,29 @@
...
+console.log("The value of my variable 'num' is: " + num);
+console.log("num is: Number");
+console.log("The value of my variable 'str' is: " + str);
+console.log("str is: String");
+console.log("The value of my variable 'arr' is: " + arr);
+console.log("arr is: Array");
+console.log("The value of my variable 'boo' is: " + boo);
+console.log("boo is: Boolean");
+
+function compareType() {
```
marcianoviereck: Wow nice function. But I think that in the assignment it is required to `console.log` it every time two variables are not the same.

> Make sure to also show a message when the variables you are comparing are not the same type.

So you could make one function where you pass two variables as `arguments` and then compare those two and if they are not the same, then log that also. And then pass different variables to that function (I can explain it in more detail if you don't fully understand what I'm saying).

marcianoviereck (changes requested):
The (new) homework looks good overall! But I do have some remarks here and there, if you can fix those then the new homework is also approved!

Week2/10_program.js
```diff
@@ -1,4 +1,17 @@
...
\ No newline at end of file
+
+
+//10.2 Can you compare infinities? (Not in Eyad's world) - does 6/0 === 10/0? How can you test this?
+//10.3 Add console.log statements to the above program
+let inf1 = 10/0;
+let inf2 = 6/0;
+
+function comparInfinities(inf1,inf2){
+        if (inf2 = Number.MAX_VALUE , inf2 = Number.MAX_VALUE)
```
marcianoviereck: This if statement will not work, you can just use a regular `if-statement` to compare the `inf1` and `inf2`

Week2/10_program.js
```diff
@@ -1,4 +1,17 @@
...
+        console.log("inf1 and inf2 are infinity")
+        else console.log("False")
+}
+comparInfinities()
```
marcianoviereck: You call the function, which requires 2 `arguments` and then pass no `arguments`, so this will not work. But, you don't have to use a function, so I would try to avoid using a function because it might be confusing right now.

Week2/8_checkTypeOfData.js
```diff
@@ -11,19 +16,16 @@ console.log("arr is: Array");
...
-}
-compareType()
+//8.5 Now compare the types of your different variables with one another.
+//8.6 Make sure to also show a message when the variables you are comparing are not the same type.
+function compareType(var1,var2) {
+    if (typeof var1 == typeof var2)
+        console.log( "this variable have the same type")
+        else console.log("The first variable assigned to %o and the second variable assigned to %o ",typeof var1 , typeof var2 )
+    }
+compareType(num,str);
```
marcianoviereck, line 28: Ah nice! This looks very good.
Just a tip, I would always add brackets (`{` and `}`) around your `if-statements` because otherwise it is easier to make mistakes.

So then it would look like this:

```
function compareType(var1,var2) {
    if (typeof var1 == typeof var2) {
        console.log( "this variable have the same type");
     } else { 
        console.log("The first variable assigned to %o and the second variable assigned to %o ",typeof var1 , typeof var2 );
      } 
}
```

marcianoviereck (changes requested):
The homework is almost approved! Only assignment 10 is not good yet! The rest looks good :)

PR #138, opened 2019-07-31
JavaScript1/week2

ghost:


Week2/homework/array.js
```diff
@@ -0,0 +1,24 @@
+//6.1 Declare an empty array. Make sure that the name you choose indicates 'plurality'.
+let animals = [];
+
+
+//6.2 Write a console.log statement that explains in words what you think the value of the array is.
+console.log('the content of my array will be about animals');
```
marcianoviereck: Can you explain here what the value of the array is right now? (I know it will be about animals, but what is the value)

Week2/homework/greeting.js
```diff
@@ -0,0 +1,8 @@
+const Dutch = 'Hallo Wereld';
+
+const French = 'Bonjour le monde';
+
+const Danish = 'Hej Verden';
+
+console.log(Dutch,fFrench,Danish);
```
marcianoviereck: I think you made a little mistake here:)

Week2/homework/round.js
```diff
@@ -0,0 +1,36 @@
...
+
+//5.5 So now we have z and a find a way to compare the two values and store the highest of the two in a new variable.
+ let h;
+ if (z > a)
+
+{
+   
+    h=z;
+
+}
```
marcianoviereck: Ah, this solution is not complete, please check it again.  Now you only check if `z` is higher than `a` and then setting `z` to the highest value. But what if `a` is higher than `z`?

Week2/homework/variable-x.js
```diff
@@ -0,0 +1,21 @@
+//3.1 First, declare your variable x (do not initialize it yet).
+
+let x;
+
+//3.2 Add a console.log statement that explains in words what you think the value of x is, like in this example:
+console.log("the value of my variable x will be: 25");
```
marcianoviereck: I know that the value will be 25, but what is the value right now?

marcianoviereck (changes requested):
The homework looks good, I do have some remarks here and there:)

Week2/homework/array.js
ghost: It supposed to be an empty array, so I think the value here is undefined!
I have changed this statement.

Week2/homework/greeting.js
ghost: yes, I have just noticed that :) and corrected it

Week2/homework/round.js
ghost: you're right, i have changed it by using Math.max().

Week2/homework/variable-x.js
ghost: The value of x here should be undefined, I will change it 👍

Week2/homework/array.js
```diff
@@ -3,7 +3,7 @@ let animals = [];
 
 
 //6.2 Write a console.log statement that explains in words what you think the value of the array is.
-console.log('the content of my array will be about animals');
+console.log('the value of my array will be zero');
```
marcianoviereck: Ah, almost correct! In your comment you said it correct, the value of `my array` is an `empty` array. So it is not zero, but an empty array:)

Week2/homework/round.js
```diff
@@ -20,17 +20,11 @@ console.log(a);
 
 
 //5.5 So now we have z and a find a way to compare the two values and store the highest of the two in a new variable.
- let h;
- if (z > a)
+let max = Math.max(a, z);
```
marcianoviereck: Nice! I would suggest using `const` though (because the value of `max` will not change anymore

marcianoviereck (changes requested):
Two small change requests, if you fix these then it is approved.

Week2/homework/array.js
ghost: done!

Week2/homework/round.js
ghost: yes definitely!

marcianoviereck (approved):
Nice! The homework is approved, nicely done!

PR #139, opened 2019-08-01
I really hope this works :)

veyselindi:


week4:JSweek2/2.js
```diff
@@ -0,0 +1,11 @@
+var Imawesome = 'I`m awesome';
+
+console.log(Imawesome);
+
+/* 
+console.log('I'm awesome');
+
+this code did not have any value above,
+and using ' caused problems. wrong format
```
nudge, line 7: Do you remember using backticks ``?
For example you could use:
````
console.log(`...`);
````

File: week4:JSweek2/1.js [hunk 2]
```diff
@@ -0,0 +1,7 @@
+var Eng = 'Hello World';
+let Tur = 'Merhaba Dunya';
+
+console.log(Eng);
+console.log(Tur);
+
+/* Console Log excercise */
```
nudge, line 7: :+1:

week4:JSweek2/3.js
```diff
@@ -0,0 +1,4 @@
+var x
+console.log("the value of my variable x will be: five");
```
nudge: `five` or `5`?

week4:JSweek2/3.js
```diff
@@ -0,0 +1,4 @@
...
+var x = (5);
```
nudge, line 3: you don't need the parentheses, `var x = 5;` also works

week4:JSweek2/4.js
```diff
@@ -0,0 +1,6 @@
+let y = 'isJSreallyHard?';
+console.log("the value of my string will be: isJSreallyHard?");
+console.log(y);
+y='ItLooksLikeThat';
+console.log("the value of my string will be: ItLooksLikeThat");
+console.log(y);
```
nudge, line 6: :+1:

week4:JSweek2/5.js
```diff
@@ -0,0 +1,5 @@
+var z = '7.25';
```
nudge: this should be initialised as a number, not a string ;)

week4:JSweek2/5.js
```diff
@@ -0,0 +1,5 @@
...
+console.log(z);
+var a = Math.ceil(7.25);
```
nudge: You could use the variable you defined above, eg `Math.ceil(z)`.
Also, did you want to use `ceil` or `round` here?

week4:JSweek2/5.js
```diff
@@ -0,0 +1,5 @@
...
+console.log(a);
+console.log(Math.max(z, a ));
```
nudge: nit: you have an extra space in the line, you can remove it to make the code neater!

week4:JSweek2/6.js
```diff
@@ -0,0 +1,7 @@
+let arr = [ ];
```
nudge: a better name might be `myFavouriteAnimals`, it is more descriptive

week4:JSweek2/6.js
```diff
@@ -0,0 +1,7 @@
...
+console.log("the value of my array will be: undefined");
+console.log(arr[0]);
+arr = ['parrot', 'fish', 'dog'];
+console.log(arr[2]);
+arr.push('baby pig');
+console.log(arr[3]);
```
nudge: You can also try log the entire array itself, eg `console.log(arr)`. What do you expect it to show?

week4:JSweek2/7.js
```diff
@@ -0,0 +1,4 @@
+let myString = "this is a test";
+console.log(myString);
+'myString'.length;
```
nudge: ^ you should delete this line

nudge:
Nice work! Added a few comments. If you can address them that would be great!

nudge:
Also for next time, remember to create a branch from `master` for your homework. This makes it easier if you want to work on multiple week's homework at the same time without your previous week's homework getting in the way.

week4:JSweek2/2.js
veyselindi: I see what I did wrong.

week4:JSweek2/3.js
veyselindi: the homework says "explains in words". I misinterpreted :)

week4:JSweek2/3.js
veyselindi: sure. I deleted the first line of the code and it still works. I do not necessarily need to declare x like that right?
in other words, is    var x = 5    enough?

week4:JSweek2/5.js
veyselindi: got rid of ' ' :)

week4:JSweek2/5.js
veyselindi: true, it does not necessarily ask me to round up. just round. need to simplify my thinking process :)

week4:JSweek2/6.js
veyselindi: I dont even know why I referred to specific items.  :))))

PR #142, opened 2019-08-03
JS1-HW2-till7

NesibeInce:


Week2/homework/index.js
```diff
@@ -0,0 +1,114 @@
+console.log(1);
```
marcianoviereck: Can you split the homework up in separate files:)?

Week2/homework/index.js
```diff
@@ -0,0 +1,114 @@
...
+let HiIndo='Hallo, dunia!'
```
marcianoviereck: Do you think it's better to use `let` or `const` here? 

And please use lowercase letter as the first letter for variables. So like this `hiIndo`. See [naming conventions](https://github.com/HackYourFuture/fundamentals/blob/master/fundamentals/naming_conventions.md)

Week2/homework/index.js
```diff
@@ -0,0 +1,114 @@
...
+//The symbol used for the string is the same with the symbol used in the syntax of the string.
+// When I changed it into sth else, i gives what i wanted to write.
+console.log(" I'm awesome!");
+
+
+console.log(3)
+let x;
+x=3;
+
+console.log("the value of my variable x will be: 3");
```
marcianoviereck: Ah, you also have to `console.log` the value of `x` before you initialize it. Can you also do that?

Week2/homework/index.js
```diff
@@ -0,0 +1,114 @@
...
+console.log(x);
+
+
+console.log(4);
+var y="I LOve Spring";
+console.log("The value of my variable y will be: I Love Spring!);
```
marcianoviereck: I don't think this will compile! Can you re-check this line?

Week2/homework/index.js
```diff
@@ -0,0 +1,114 @@
...
+console.log(3)
+let x;
+x=3;
+
+console.log("the value of my variable x will be: 3");
+console.log(x);
+
+
+console.log(4);
+var y="I LOve Spring";
```
marcianoviereck: Please don't use `var`. Use `let` or `const` instead

Week2/homework/index.js
```diff
@@ -0,0 +1,114 @@
...
+console.log("The value of my variable y will be: I Love Spring!);
+
+var y="The Rain Is Beautiful!";
```
marcianoviereck: I don't think this will work, because you already defined `y` at line 40. Please recheck this line

Week2/homework/index.js
```diff
@@ -0,0 +1,114 @@
...
+console.log("the value of my variable y will be: The Rain Is Beautiful!");
+console.log(y);
+
+
+console.log('5th question');
+var z = 7.25;
+console.log(z);
+var a = Math.round(7.25);
```
marcianoviereck: Can you replace `var` with `let` or `const`

Week2/homework/index.js
```diff
@@ -0,0 +1,114 @@
...
+console.log(a);
+
+if (a > z) {let c = a;
+console.log(c)} else {let c = z; console.log(c)};
```
marcianoviereck: Ahh, yeah this works technically! But an easier solution is using [Math.max](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max)

Week2/homework/index.js
```diff
@@ -0,0 +1,114 @@
...
+console.log(animals);
+animals[0] = 'dog';
+animals[1] = 'cat';
+animals[2] = 'lion';
+animals[3] = 'cow';
+animals[4] = 'bird';
+console.log(animals);
+animals[5] = '
baby pig';
+console.log(animals);
```
marcianoviereck: Ahh, yeah this all works, but an easier (and cleaner) solution is by using [push](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push)

marcianoviereck (changes requested):
The homework looks pretty nice! I do have quite some remarks here and there:)

Week2/homework/index.js
NesibeInce: Hi,@marcianoviereck!
Yes, i think it is better to use const here.
i have made the necessary changes. i hope it is ok now.
Thank you for your attention.

Week2/homework/index.js
NesibeInce: Hi,
i knew about  this actually. 
i don't know why i did it so.
i have corrected.
Thank you .

Week2/homework/index.js
NesibeInce: Hi,
Thank you for the enlightening.
I made a new .js files.
i gave them explanatory names.
and wrote short comments for each of the questions. 

Regards

Week2/HW2/10.multipleTypesArray.js
```diff
@@ -0,0 +1,9 @@
+// store multiple types in the same array
+
+let arr = ['Amsterdam', 1,'Rotterdam', 2, true ];
```
marcianoviereck, line 3: Do you think it's better to use `let` or `const` here?

Week2/HW2/10.multipleTypesArray.js
```diff
@@ -0,0 +1,9 @@
...
+console.log (arr);
+
+let x;
+
+x = 6/0 === 10/0;
```
marcianoviereck, line 8: Nice! you can also do this in one line: `const x = 6/0 === 10/0` :)

Week2/HW2/3.initializingStatement.js
```diff
@@ -0,0 +1,8 @@
+// Adding a statement, initializing, and explaining it through console.log 
+
+let x;
+console.log("the value of my variable x will be a number");
```
marcianoviereck, line 4: Ah, do you think it's a number `now` ? I know it `will` be a number, but what is the value right now?

Week2/HW2/4.givingVariables.js
```diff
@@ -0,0 +1,9 @@
+// giving two different variables to y.
+
+let y = 'Í love spring.';
+console.log("The value of my variable y will be: a string");
```
marcianoviereck, line 4: Same question here:)

Week2/HW2/4.givingVariables.js
```diff
@@ -0,0 +1,9 @@
...
+console.log (y);
+
+let x = 'Flowers are beautiful.';
+console.log("the value of my variable y will be: a string");
```
marcianoviereck, line 8: And will this be the value? I know the `type` will be a string, but what is the value?

Week2/HW2/5.roundNumber.js
```diff
@@ -0,0 +1,8 @@
+// rounding a number and printing out the highest number
+let z = 7.25;
+console.log(z);
+let a = Math.max(7.25);
+console.log(a);
+
+if (a > z) {let c = a;
```
marcianoviereck, line 7: You don't have to do this, `Math.max` is already good enough:) So this code is not needed.

Week2/HW2/6.array.js
```diff
@@ -0,0 +1,13 @@
...
+console.log('for now,it is an empty array, but later there will be names of animals.');
+console.log(animals);
+animals.push('dog');
+animals.push('cat');
+animals.push('lion');
+animals.push('cow');
+animals.push('bird');
+console.log(animals);
+animals.push('
baby pig');
```
marcianoviereck, line 12: nicenice!

Week2/HW2/8.sameType.js
```diff
@@ -0,0 +1,47 @@
...
+let t = false;
+console.log (t);
+
+if ( typeof x === typeof y) {
+    console.log (`${x} and  ${y} are in the same type`)
+} else { 
+    console.log (`${x} and  ${y} are not in the same type`)
+};
+
+if ( typeof x === typeof y) {
```
marcianoviereck, line 19: Ah, I think you forgot to change `x` and `y`. Because you do have all the `comparisons` but you have `x` and `y` everywhere:)

Week2/HW2/9.remainder.js
```diff
@@ -0,0 +1,17 @@
...
+let p = 10;
+p = p % 4; // 2
+console.log(p);
+
+let o = 11;
+o = o % 2; // 1
+console.log(o);
+
+let x = 7;
+x = x % 3; // 1
```
marcianoviereck, line 16: Nicenice! Looks good!

marcianoviereck (changes requested):
The homework looks a lot better! I do have some small feedback though:) But keep it up!

PR #146, opened 2019-08-13
js1-week3-homework

mehmetekinci:


Week3/script.js
```diff
@@ -0,0 +1,273 @@
+/* *** String! *** */
+
+// 1.1
```
marcianoviereck, line 3: Could you split up your homework in separate files?

See this:

> For all the following exercises create a new .js file.

Week3/script.js
```diff
@@ -0,0 +1,273 @@
...
+
+console.log(myString);
+
+// 1.2
+
+console.log(myString.length);
+
+// 1.3 & 1.4
+
+console.log(myString.replace(/,/g, " "));
```
marcianoviereck: Nice solution!

Week3/script.js
```diff
@@ -0,0 +1,273 @@
...
+
+// 2.9
+
+console.log(favoriteAnimals.indexOf("meerkat"));
+
+// 2.10
+
+console.log("The item you are looking for is at index: 1");
+
+        /* *** More JavaScript *** */
```
marcianoviereck: I think you might have forgotten to remove this comment?

Week3/script.js
```diff
@@ -0,0 +1,273 @@
...
+function threeArguments(a, b, c) {
+  var sum = a + b + c;
+
+  return sum;
+}
+
+        // 2
+
+function colorCar(car) {
+  return car;
```
marcianoviereck: Looks good! But I think the function itself should `console.log("a red car")`. Now the function only returns the variable to you pass into it. But it should use the value you pass into it to `console.log` the correct value.

Week3/script.js
```diff
@@ -0,0 +1,273 @@
...
+let car = "a red car";
+
+console.log(colorCar(car));
+
+        // 3
+
+const person = {
+  name: "Mehmet Ekinci",
+  gender: "M",
+  adres() {
```
marcianoviereck: I don't think this works as a property, did you try to run this part of the code?

Week3/script.js
```diff
@@ -0,0 +1,273 @@
...
+    console.log("Koemeershoeve");
+  }
+};
+console.log(person);
+
+        // 4
+
+function vehicleType(color,code) {
```
marcianoviereck: looks nice!

Week3/script.js
```diff
@@ -0,0 +1,273 @@
...
+    type = " car";
+  } if (code === 2) {
+    type = " motorbike";
+  } if (age >= 2) {
+    condition = " used";
+  } if ((age >= 0) && (age <= 1)) {
+    condition = " new";
+  }
+  
+  console.log("a " + color + condition + type);
```
marcianoviereck: This looks nice, but I would recommend to add some more spaces and use `else if`. Like this:

```
  if (code === 1) {
    type = " car";
  } else if (code === 2) {
    type = " motorbike";
  } 
  if ((age >= 0) && (age <= 1)) {
    condition = " new";
  } else if (age >= 2) {
    condition = " used";
  } 
```

This makes it easier to read (and the logic is easier to follow:))

Week3/script.js
```diff
@@ -0,0 +1,273 @@
...
+
+}
+vehicle("blue",1,5);
+
+        // 7
+
+let vehicles = ["motorbike", "caravan", "bike", "car", ]
```
marcianoviereck: Does it work with the `,` at the end? (and don't forget the `;` ;))

Week3/script.js
```diff
@@ -0,0 +1,273 @@
...
+
+        // 8
+
+console.log(vehicles[2]);
+
+        // 9
+
+function newVehicle(color,code,age) {
```
marcianoviereck: I know it's not clear from the assignment, but the `code` should be used to get the value from the array.

Week3/script.js
```diff
@@ -0,0 +1,273 @@
...
+        // 10
+let advertisement = "Amazing Joe's Garage, we service "
+function reAdvertisement(advertisement, vehicles){
+
+  for( let i = 0; i < vehicles.length ;i++){
+    const lastElement = vehicles[vehicles.length-1];
+    const beforLastElement = vehicles[vehicles.length-2];
+    const currentElement = vehicles[i];
+
+    if (currentElement === lastElement) {
```
marcianoviereck: Ah nice!

This looks good, you can also use the `indexes` though and then compare the `index` numbers. So then it would look like this:

```
for( let i = 0; i < vehicles.length ;i++){
    if (i=== vehicles.length - 1) {
     advertisement = advertisement + currentElement + ".";
    } else if (i=== vehicles.length - 2) {
      advertisement = advertisement + currentElement + " and ";
    } else {
      advertisement = advertisement + currentElement + ", "
    }
```

Week3/script.js
```diff
@@ -0,0 +1,273 @@
...
+console.log(reAdvertisement(advertisement,vehicles));
+
+        // 11
+
+// If you add more vehicle to array, you will be add to advertisement.
+function newFunction(){
+vehicles.push("truck");
+ return console.log(reAdvertisement(advertisement,vehicles))
+}
+newFunction();
```
marcianoviereck, line 182: nice!

Week3/script.js
```diff
@@ -0,0 +1,273 @@
...
+
+        // 12
+
+let newObject = {};
+
+        // 13
+
+let hackYourFuture = new Object();
+
+hackYourFuture.teachers = "Philipp, Rob, Unmesh, Bonan, Yash";
```
marcianoviereck: It's good practice to create an `array` object for multiple values, instead of a string with `commas`.
So could you change this into an array:)?

Week3/script.js
```diff
@@ -0,0 +1,273 @@
...
+
+console.log(hackYourFuture);
+
+
+
+        // 14
+
+hackYourFuture.langs = "HTML, CSS, CLI, GIT-Bash, JS";
```
marcianoviereck: Same here

Week3/script.js
```diff
@@ -0,0 +1,273 @@
...
+
+const o1 = { foo: "bar" };
+const o2 = { foo: "bar" };
+const o3 = o2;
+console.log(o2.foo)
+console.log(o3)
+
+console.log(o1 == o2);  // false
+console.log(o2 == o3);  // true
+console.log(o3 == 01);  // false
```
marcianoviereck: I think you wrote `01` instead of `o1` (zero instead of o)

marcianoviereck: Can you also change that?

marcianoviereck (changes requested):
The homework looks really nice overall:) I do have some review comments here and there. But overall the functions are really nice, keep it up!

yash-kapila:
@mehmetekinci could you please have a look at the review comments and try to implement the requested changes? 🙂

Week3/script.js
mehmetekinci: oo I'm sorry for that. Should I do new pull request?

mehmetekinci:
@marcianoviereck  Thank you for your help and correction. I corrected all.

mehmetekinci:
@yash-kapila I did it. thank you for reminding :)

Week3/script.js
mehmetekinci: @marcianoviereck  Thank you :)  I actually knew that way but I tried to make from different way. Which way is the best?

Week3/script.js
mehmetekinci: after your feedback I tried another thing. I hope it works.

Week3/script.js
marcianoviereck: Ah, both ways are fine:) I think the one I suggested is a bit easier (and less error-prone with the comparisons). So it's a `safer` option to compare numbers instead of whole objects:)

Week3/script.js
```diff
@@ -204,15 +203,15 @@ let newObject = {};
...
+hackYourFuture.teachers = ['Philipp', 'Rob', 'Unmesh', 'Bonan', 'Yash'];
 
 console.log(hackYourFuture);
 
 
 
         // 14
 
-hackYourFuture.langs = "HTML, CSS, CLI, GIT-Bash, JS";
+hackYourFuture.langs = ['HTML', 'CSS', 'CLI', 'GIT-Bash', 'JS'];
```
marcianoviereck: Ah, i made a mistake in my first homework-review! I said that using an `array` is a better option and I think it is a better option. But now it's not clear which teacher has taught which class(es). Do you know how you can change these code so that it is clearer which teacher taught which classes(Hint: you can use object properties and arrays.

If you don't understand what I mean, please let me know! Then I will try to clarify more.

Week3/script.js
marcianoviereck: No, no problem!

Week3/script.js
```diff
@@ -72,24 +72,22 @@ function threeArguments(a, b, c) {
 
         // 2
 
-function colorCar(car) {
-  return car;
+function colorCar() {
```
marcianoviereck: Ah, now the function `always` returns `a red car`. Can you change the code so that the color that it will print out can be passed in the function. So something like this:

```
function  colorCar(color) {
 //do something with the color to print "a _ car"
}
```
```

File: Week3/script.js [hunk 17]
```diff
@@ -72,24 +72,22 @@ function threeArguments(a, b, c) {
...
-  gender: "M",
-  adres() {
-    console.log("Koemeershoeve");
-  }
+  name: 'Mehmet',
 };
-console.log(person);
+
+function personInfo() {
+  return `His ${Object.keys(person)} is ${person.name}`;
```
marcianoviereck: Actually I think just `console.log(person)` is fine. There is a nicer way of doing it, but for now this is fine:).

To read more about ways to do this, you can read [this](https://zellwk.com/blog/looping-through-js-objects/) (and if you want, you can change your code;)

Week3/script.js
```diff
@@ -120,11 +118,12 @@ function vehicle(color,code,age) {
 
   if (code === 1) {
     type = " car";
-  } if (code === 2) {
+  } els if (code === 2) {
```
marcianoviereck: Ah almost, there is a typo here. Can you take a look at this again?

Week3/script.js
```diff
@@ -135,7 +134,7 @@ vehicle("blue",1,5);
 
         // 7
 
-let vehicles = ["motorbike", "caravan", "bike", "car", ]
+let vehicles = [" motorbike", " caravan", " bike", " car"];
```
marcianoviereck: Ah, why did you add the spaces here? I think it was fine before (except for the `,`)

Week3/script.js
```diff
@@ -146,13 +145,13 @@ console.log(vehicles[2]);
...
+    type = vehicles[0];
   } if (code === 2) {
-    type = " caravan";
+    type = vehicles[1];
   }  if (code === 3) {
-    type = " bike";
+    type = vehicles[2];
   } if (code === 4) {
-    type = " car";
+    type = vehicles[3];
```
marcianoviereck: Ah, do you see a `pattern` here? Now you changed the code to this:

```
  if (code === 1) {
    type = vehicles[0];
  }
  if (code === 2) {
    type = vehicles[1];
}
..
```

The `index` of the array that you use, is always `1` value lower than the value checked in your `if-statement`. Do you think you can change the code so that you can use the value in `code` and use it as an `index` for the list?

marcianoviereck (changes requested):
It's starting to look better! I do have some change requests:) If anything is not clear, feel free to slack me!

mehmetekinci:
@marcianoviereck I trried to correct it again. I hope,I succeed it :)

Week3/script.js
```diff
@@ -54,192 +52,201 @@ console.log(favoriteAnimals);
...
+const personInfo = {
   name: 'Mehmet',
 };
 
-function personInfo() {
-  return `His ${Object.keys(person)} is ${person.name}`;
+const entries = Object.entries(personInfo);
+
+for (const [personInfo, name] of entries) {
+  console.log(`His ${personInfo} is ${name}`);
```
marcianoviereck, line 86: Nice!

Week3/script.js
```diff
@@ -54,192 +52,201 @@ console.log(favoriteAnimals);
...
-  } if (code === 4) {
-    type = vehicles[3];
-  } if (age >=2) {
-    condition = " used";
-  } if ((age >= 0) && (age <=1)) {
-    condition = " new";
+    type = vehicles[code - 1];
+  }
+  if (code === 2) {
+    type = vehicles[code - 1];
```
marcianoviereck: Ah, it's looking better! You can also do this:

```
 type = vehicles[code - 1];
```

Then you don't need all those 
```
 if (code === 2) {
}
if(code === 3) {
}
```

etc. So could you change that aswell?

Week3/script.js
```diff
@@ -54,192 +52,201 @@ console.log(favoriteAnimals);
...
+  }
+  if (code === 3) {
+    type = vehicles[code - 1];
+  }
+  if (code === 4) {
+    type = vehicles[code - 1];
+  }
+  if (age >= 2) {
```
marcianoviereck, line 147: Nice! This looks nice

Week3/script.js
```diff
@@ -54,192 +52,201 @@ console.log(favoriteAnimals);
...
+
+// 10
+let advertisement = "Amazing Joe's Garage, we service ";
+function reAdvertisement(advertisement, vehicles) {
+  for (let i = 0; i < vehicles.length; i++) {
+    // const lastElement = vehicles[vehicles.length - 1];
+    // const beforLastElement = vehicles[vehicles.length - 2];
+    // const currentElement = vehicles[i];
+
+    if (i === vehicles.length - 1) {
```
marcianoviereck, line 163: Ah nice! This looks very good, I think you forgot to remove the `commented` code on line 171, 172 and 173

Week3/script.js
```diff
@@ -54,192 +52,201 @@ console.log(favoriteAnimals);
...
 console.log(hackYourFuture);
 
+// 14
 
-
-        // 14
-
-hackYourFuture.langs = ['HTML', 'CSS', 'CLI', 'GIT-Bash', 'JS'];
+hackYourFuture.teacher1.language = ['HTML'];
+hackYourFuture.teacher2.language = ['CSS'];
```
marcianoviereck, line 208: Ah very good! 😎

Week3/script.js
```diff
@@ -0,0 +1,277 @@
...
+console.log(o1 == o2); // true
+console.log(o2 == o3); // false
+console.log(o3 == o1); // false
+console.log(o2 == o1); // false
+
+// when we change o2 whit o3, giving syntax error
+
+// Does the order that you assign (o3 = o2 or o2 = o3) matter?
+
+// sure. when you use o2 = o3, o2 declares twice. giving syntax error.
```
marcianoviereck: Ah could you show this in code? I understand that you explained it in comments, but could you explain it using code?

marcianoviereck (changes requested):
Most of the changes look really nice! I do have some small feedback left (maybe some of it I forgot the first time, sorry for that)

mehmetekinci:
@marcianoviereck hi Marciano I am sorry for late respond. I made some changes. I learn a loth with your feedbacks thank you for all.

Week3/script.js
```diff
@@ -250,24 +257,21 @@ const o1 = o2;
...
 
 // when we change o2 whit o3, giving syntax error
 
 // Does the order that you assign (o3 = o2 or o2 = o3) matter?
 
-// sure. when you use o2 = o3, o2 declares twice. giving syntax error.  
-
-
-
+// sure. when you use o2 = o3, o2 declares twice. giving syntax error.
```
marcianoviereck, line 257: Ah yeah true, but in the assignment the code is as follows:

```
let o1 = { foo: "bar" };
let o2 = { foo: "bar" };
let o3 = o2;
```

And you did `const` then it's true that an error will occur :) but if you use `let`, will the order that you assign o3 = o2 or o2 = o3 matter?

And if you change o2, will it change o3? 
And if you change o3, will it change o2?

marcianoviereck (changes requested):
I only have one more change request :) the rest looks good!

PR #147, opened 2019-08-14
JavaScript1 Week3 Homework

ihk2018:


Week3/advertisement.js
```diff
@@ -0,0 +1,13 @@
...
+let advertisedVehicles = '';
+
+for (let i = 0; i < vehicles.length; i++) {
+  if (i == vehicles.length - 1) {
+    advertisedVehicles += 'and ' + vehicles[i] + 's.';
+  } else {
+    advertisedVehicles += vehicles[i] + 's, ';
+  }
+}
+console.log(`Amazing HappyMan's Garage, we service ${advertisedVehicles}`);
```
marcianoviereck: Nice! This looks good:)

Week3/changeVehicleFunction.js
```diff
@@ -0,0 +1,25 @@
+//9. Change the function vehicle to use the list of question 7. So that vehicle("green", 3, 1) prints "a green new bike".
+
+function myVehicles(color, code, age) {
+  let vehicles = ['motorbike', 'caravan', 'bike', 'truck'];
+  let myChosenVehicle;
+  let myVehicleType;
+  let i;
+
+  for (i = 0; i < vehicles.length; i++) {
```
marcianoviereck: neat! You can also do this easier without using a `for-loop`:

```
if(code >= 0 && code < vehicles.length) {
  myChosenVehicle = vehicles[code];
}
```

Week3/comparisonOfArrays.js
```diff
@@ -0,0 +1,48 @@
...
+console.log(myComparison(z, y));
+
+// What do you think will happen with x == y, x === y and z == y and z == x? Prove it!
+
+x == y; // false because their types are objects and they reference different objects
+x === y; // false because as x==y then they can never be x===y
+z === y; // true, because they reference the same objects
+z == x; // false, because they reference different objects
+
+function myComparison(a, b) {
```
marcianoviereck, line 32: Did you accidentally create this function twice?

Week3/comparisonOfArrays.js
```diff
@@ -0,0 +1,48 @@
...
+  }
+}
+
+myComparison(x, y); //false
+myComparison(x, z); //false
+myComparison(z, y); // true
+
+console.log(myComparison(x, y));
+console.log(myComparison(x, z));
+console.log(myComparison(z, y));
```
marcianoviereck, line 48: Same with this part

Week3/functionVehicle.js
```diff
@@ -0,0 +1,19 @@
+// 6. Create a function called vehicle, like before, but takes another parameter called age, so that vehicle("blue", 1, 5) prints 'a blue used car'
+
+function vehicle(color, code, age) {
+  if (age >= 1 && code === 1) {
```
marcianoviereck: Looks good, you can do this easier by splitting the `if-statements`.

You can make one to determine the `age`. So you could do this:

```
if (age >= 1) {
 ageText = `new`;
} else {
 ageText = `used`
}
```

And the  same for the rest, if you understand what I mean. But this looks good anyways, but it's just a suggestion

Week3/objectIncludingHyfTeachers.js
```diff
@@ -0,0 +1,7 @@
+// 13. Create an object that contains the teachers that you have had so far for the different modules.
+
+const hyf = new Object();
+
+hyf.Teachers = 'Philipp, Rob, Unmesh, Bonan, Yash';
```
marcianoviereck: Ah, usually it's a good practice to use an `array` for `multiple-values`. So could you also use that here?

Week3/objectIncludingLanguages.js
```diff
@@ -0,0 +1,11 @@
+// 14.Add a property to the object you just created that contains the languages that they have taught you.
+
+const hyf = new Object();
+
+hyf.Teachers = 'Philipp, Rob, Unmesh, Bonan, Yash';
+
+console.log(hyf);
+
+hyf.Languages = 'HTML, CSS, CLI, GIT, JS1';
```
marcianoviereck: same here

Week3/string.js
```diff
@@ -0,0 +1,27 @@
...
+
+console.log(myString.length);
+
+// Q:The commas make that the sentence is quite hard to read. Find a way to remove the commas from the string and replace them with spaces.
+
+// Solution
+
+const myString = 'hello,this,is,a,difficult,to,read,sentence';
+
+let regex = /,/gi;
```
marcianoviereck, line 25: Clear! But I think you've redefined `myString` and because you already defined it on line `5` (and it's a `const`) so I don't think this will work properly

marcianoviereck (changes requested):
The homework looks really good! I do have some feedback here and there, but keep it up :)

Week3/advertisement.js
ihk2018: Dear Marciano, thank you for the feedback. I'll review and make the needful changes. 

Regards,

Week3/moreVehicleToAdvertisementList.js
```diff
@@ -0,0 +1,17 @@
...
+
+vehicles = ['motorbike', 'caravan', 'bike', 'truck'];
+console.log(vehicles);
+vehicles.push('car');
+console.log(vehicles);
+
+let advertisedVehicles = '';
+
+for (let i = 0; i < vehicles.length; i++) {
+  if (i == vehicles.length - 1) {
```
marcianoviereck: Ah, I forgot to check this one. It looks good but I think now it will result in:

`Amazing HappyMan's Garage, we service motorbikes, caravans, bikes, and trucks`.

But the `,` before the `and` is not correct:) can you make it so that the `,` is not added before the `and`?

marcianoviereck (changes requested):
Forgot one thing

Week3/changeVehicleFunction.js
ihk2018: I changed the code as you requested and it works perfectly :)

Week3/comparisonOfArrays.js
ihk2018: Actually, I did not loaded the function accidentally , i think the question repeats itself so i thought its better to add the solution one more time .

Week3/functionVehicle.js
ihk2018: I changed it, its better now, thanks.

Week3/string.js
ihk2018: You are completely rigth, i change the string name , so it works now.

Week3/moreVehicleToAdvertisementList.js
ihk2018: I found a solution for that as below; and it works. 


function newAdvertisement(advertisedVehicles, vehicles) {
  for (let i = 0; i < vehicles.length; i++) {
    if (i == vehicles.length - 1) {
      advertisedVehicles += 'and ' + vehicles[i] + 's.';
    } else if (i == vehicles.length - 2) {
      advertisedVehicles += vehicles[i] + 's ';
    } else {
      advertisedVehicles += vehicles[i] + 's, ';
    }
  }
  return advertisedVehicles;
}

console.log(newAdvertisement(advertisedVehicles, vehicles));

Week3/advertisement.js
```diff
@@ -1,13 +1,20 @@
...
 
-let advertisedVehicles = '';
+let advertisedVehicles = "Amazing HappyMan's Garage, we service ";
 
-for (let i = 0; i < vehicles.length; i++) {
-  if (i == vehicles.length - 1) {
-    advertisedVehicles += 'and ' + vehicles[i] + 's.';
-  } else {
-    advertisedVehicles += vehicles[i] + 's, ';
+function newAdvertisement(advertisedVehicles, vehicles) {
```
marcianoviereck, line 7: Nice!

Week3/functionVehicle.js
```diff
@@ -1,19 +1,25 @@
...
+  let ageText;
+
+  if (age >= 1) {
+    ageText = `used`;
   } else {
-    return `Please select age of the vehicle `;
+    ageText = `new`;
   }
+
+  if (code === 1) {
```
marcianoviereck, line 13: Wow great! The code also looks a lot cleaner!

Week3/moreVehicleToAdvertisementList.js
```diff
@@ -1,17 +1,20 @@
...
+    } else {
+      advertisedVehicles += vehicles[i] + 's, ';
+    }
   }
+  return advertisedVehicles;
 }
-console.log(`Amazing HappyMan's Garage, we service ${advertisedVehicles}`);
+
+console.log(vehicles);
+// vehicles.push('car');
```
marcianoviereck: Ah you still have this line commented out, why is that?

(And I think you forgot to copy paste your new function that you made in `advertisement.js` file)

Could you take another look at this? When this is okay then the homework is approved:)!

Week3/objectIncludingHyfTeachers.js
```diff
@@ -1,7 +1,19 @@
 // 13. Create an object that contains the teachers that you have had so far for the different modules.
 
-const hyf = new Object();
-
-hyf.Teachers = 'Philipp, Rob, Unmesh, Bonan, Yash';
-
-console.log(hyf);
+const teachers = {
+  hyfTeacher1: {
```
marcianoviereck, line 4: Ah good, an `array` would also have worked :) But this is fine

marcianoviereck (changes requested):
The new changes are really nice! I only have one more change request, then the homework is approved:)

Week3/moreVehicleToAdvertisementList.js
ihk2018: I forgot to modify it :) But I corrected it now.

marcianoviereck (approved):
Nice! Good job:) homework approved!

PR #152, opened 2019-08-14
Add Js1week3

FarukTercan:


Week3/1. Strings.js
```diff
@@ -0,0 +1,16 @@
...
+//  Add the string to your file and log it.
+console.log(myString); // hello,this,is,a,difficult,to,read,sentence
+
+
+//  Log the length of myString.
+console.log(myString.length); // 42
+
+//  Replace commas with spaces ????
+let myString2 = myString.replace(/,/g , ' ');
+console.log(myString2);  // hello this is a difficult to read sentence
```
marcianoviereck, line 14: Nice! Do you think its better to use `let` here, or `const`?

Week3/More JavaScript.js
```diff
@@ -0,0 +1,180 @@
+
+/**************************** More JavaScript ***************************** */
+
+// 1. Create a function that takes 3 arguments and returns the sum of the these arguments.
+
+function f1(a,b,c) {
```
marcianoviereck: Nice! But could you give the function a more descriptive name? `f1` doesn't really say much about what the function does.

Week3/More JavaScript.js
```diff
@@ -0,0 +1,180 @@
...
+
+const bestDriver = {
+    firstName:"Mika", 
+    lastName:"Hakkinen", 
+    teamName:"McLaren", 
+    age:50};
+    
+function printObj(obj) {
+    for (const key in obj) {
+        console.log(`His ${key} : ${obj[key]}`)
```
marcianoviereck, line 30: Clean solution!

Week3/More JavaScript.js
```diff
@@ -0,0 +1,180 @@
...
+
+
+// 4. Create a function named vehicleType that receives a color, and a code, 1 for car, 2 for motorbike. 
+    // And prints 'a blue motorbike' for example when called as vehicleType("blue", 2)
+
+let vehicleCategory = {
+    1 : "car",
+    2 : "motorbike", 
+};
+    function vehicleType(color,category) {
```
marcianoviereck: Ah, clever! You could also use an array, but this is also fine! But what if you do `vehicleType("green", 3)` ? What happens then?

Week3/More JavaScript.js
```diff
@@ -0,0 +1,180 @@
...
+// 6. Create a function called vehicle, like before, but takes another parameter called age, 
+    //so that vehicle("blue", 1, 5) prints 'a blue used car'
+
+function vehicle(color, category, age) {
+    if (age >= 1) {
+        age = "used";
+    } else {
+        age = "new";
+    }
+    console.log("a " + color + " " + age + " " + vehicleCategory[category]);
```
marcianoviereck: Same question here as before, what if I do this: `vehicle("green", 5, 5)`?

Week3/More JavaScript.js
```diff
@@ -0,0 +1,180 @@
...
+
+// 9. Change the function vehicle to use the list of question 7. So that vehicle("green", 3, 1) prints "a green new bike".
+
+function newVehicle(color, category, age) {
+    if (age >= 4) {
+        age = "used";
+    } else {
+        age = "new";
+    }
+    console.log("a " + color + " " + age + " " + listOfVehicle[category]);
```
marcianoviereck: Again the same question:)

Week3/More JavaScript.js
```diff
@@ -0,0 +1,180 @@
...
+
+ 
+// 10. Use the list of vehicles to write an advertisement. So that it prints something like: "Amazing Joe's Garage, we service cars, motorbikes, caravans and bikes.". (Hint: use a for loop.)
+
+listOfVehicle = ["car", "motorbike", "caravan", "bike"];
+
+let nameOfVehicle = "";
+
+for (i = 1; i < listOfVehicle.length; i++) {
+  if (i < listOfVehicle.length - 1) {
```
marcianoviereck, line 128: Ah, you almost got it, but now the result will be:

`Amazing Joe's Garage, we service cars, motorbikes, caravans, and bikes.`.

But the `,` before the `and` is not correct. There should not be a `,` before the `and`. So could you change that:)? The rest looks nice!

Week3/More JavaScript.js
```diff
@@ -0,0 +1,180 @@
...
+
+listOfVehicle = ["car", "motorbike", "caravan", "bike"];
+
+listOfVehicle.push("jeep");
+
+nameOfVehicle = '';
+
+for (i = 1; i < listOfVehicle.length; i++) {
+  if (i < listOfVehicle.length - 1) {
+    nameOfVehicle += ", " + listOfVehicle[i] + "s";
```
marcianoviereck: Same here ofcourse

Week3/More JavaScript.js
```diff
@@ -0,0 +1,180 @@
...
+// 12. Create an empty object.
+
+    const class23 = new Object ();
+
+// 13. Create an object that contains the teachers that you have had so far for the different modules.
+// 14. Add a property to the object you just created that contains the languages that they have taught you.
+
+    class23.teachers = 'Philipp, Rob, Unmesh & Bonan, Yash';
+
+    class23.languages = 'Html, CSS, Git, JavaScript1'
```
marcianoviereck: Nicnice, could you use an `array` instead of a long `string`? Usually it's good practice to use an `array` for `multiple values`

Week3/More JavaScript.js
```diff
@@ -0,0 +1,180 @@
...
+    const o3 = o2;
+
+// Show that changing o2 changes o3 (or not) and changing o1 changes o3(or not).
+    
+    /* Changing o2 changes o3 since o3 is referenced to value of o2.
+    But changing o1 doesn`t change o3 since they are not correlated.*/
+
+// Does the order that you assign (o3 = o2 or o2 = o3) matter?
+    
+    /* It does matter since o2 is a constant and cannot be reassigned. */
```
marcianoviereck, line 212: Ah, could you also show this by `console.log`'ing the results of making the changes?

marcianoviereck (changes requested):
The homework looks really nice! There are some clean and cool things that you have done! But I do have some feedback here and there:)

yash-kapila:
@FarukTercan could you please have a look at the review comments and apply the requested changes? 🙂

Week3/1. Strings.js
FarukTercan: Hi Marciano, first of all, I am deeply sorry for my late response.  I tried to make all implements according to your feedback.
I think, using let is better here since it allows to change in string in case of re-assign needed.

Week3/More JavaScript.js
FarukTercan: Thanks :) I changed it

Week3/More JavaScript.js
FarukTercan: I added console.log for the results

Week3/More JavaScript.js
FarukTercan: I changed f1 with sum. I think it is more suitable

Week3/More JavaScript.js
FarukTercan: I made some changes and added console.log for the undefined category inputs.
Also implemented the same solution for the similar questions below

Week3/More JavaScript.js
FarukTercan: I fixed the problem. (I fixed also for a similar issue in the next question)

Week3/More JavaScript.js
FarukTercan: Thanks for the tip. I revised it with array.

Week3/More JavaScript.js
FarukTercan: I added console.log for the results.

Week3/1. Strings.js
FarukTercan: Hi Marciano, first of all, I am deeply sorry for my late response.  I tried to make all implements according to your feedback.
I think, using let is better here since it allows to change in string in case of re-assign needed.

Week3/More JavaScript.js
```diff
@@ -42,11 +42,28 @@ let vehicleCategory = {
...
-     
-vehicleType("blue", 2);
+vehicleType('blue', 2);
+vehicleType('green', 3);
+
+// I could have set the function vehicleType also with array like this. 
+
+/* let vehicleCategory = ['car', 'motorbike'];
+function vehicleType(color, category) {
+  if (category <= 2) {
```
marcianoviereck, line 59: Ah, if you do it this way, you can use `category` <= `vehicleCategory.length` instead of `category` <= `2` :)

Week3/More JavaScript.js
```diff
@@ -42,11 +42,28 @@ let vehicleCategory = {
     1 : "car",
     2 : "motorbike", 
 };
-    function vehicleType(color,category) {
-        console.log("a " + color + " " + vehicleCategory[category] );        
+    function vehicleType(color, category) {
```
marcianoviereck: Ah, I think that you have too many `tabs` before your function. If you change it to this it will be easier to read:

```
function vehicleType(color, category) {
  if (category <= 2) {
    console.log('a ' + color + ' ' + vehicleCategory[category]);
  } else {
    console.log('Please enter valid');
  }
}
```

Week3/More JavaScript.js
```diff
@@ -58,15 +75,20 @@ console.log(3 === 3 ? "yes" : "no");
...
     }
-    console.log("a " + color + " " + age + " " + vehicleCategory[category]);
-} 
-    
-vehicle("blue", 1, 5);
+    console.log('a ' + color + ' ' + age + ' ' + vehicleCategory[category]);
+  }
+}
+
+vehicle('blue', 1, 5);
```
marcianoviereck, line 90: Will it work if you do:
`vehicle('blue', 3, 5);`
?

Week3/More JavaScript.js
```diff
@@ -79,15 +101,19 @@ console.log(listOfVehicle[2]);
 // 9. Change the function vehicle to use the list of question 7. So that vehicle("green", 3, 1) prints "a green new bike".
 
 function newVehicle(color, category, age) {
-    if (age >= 4) {
-        age = "used";
+  if (category > 5) {
```
marcianoviereck: Here you could also use `listOfVehicle.length`. You can read more about it here: https://developer.mozilla.org/nl/docs/Web/JavaScript/Reference/Global_Objects/Array

Week3/More JavaScript.js
```diff
@@ -132,11 +160,16 @@ console.log("Amazing Joe's Garage, we service " + listOfVehicle[0] + "s" + nameO
 // 13. Create an object that contains the teachers that you have had so far for the different modules.
 // 14. Add a property to the object you just created that contains the languages that they have taught you.
 
-    class23.teachers = 'Philipp, Rob, Unmesh & Bonan, Yash';
+class23.teachers = ['Philipp', 'Rob', 'Unmesh & Bonan', 'Yash'];
 
-    class23.languages = 'Html, CSS, Git, JavaScript1'
- 
-    console.log(class23); // {teachers: "Philipp, Rob, Unmesh & Bonan, Yash", languages: "Html, CSS, Git, JavaScript1"}
+class23.languages = ['Html', 'CSS', 'Git', 'JavaScript1'];
```
marcianoviereck: Ah, it is better now! But I actually made a mistake on my first review of the code. 
It is better that `teachers` and `languages` are an array, but now it is difficult to know which teacher taught which language?

So for example, if `Philipp` taught `HTML` and `CSS`, now there is no way to know that. Hint: You could use an array with objects in it, with arrays as properties, like this:
```
const objects = [
{
 name: 'someone',
 hobbies: ['first', 'second']
},
{
 name: 'someone2',
 hobbies: ['first', 'second']
},
]
```

Do you think you could change your code to look more like that?

marcianoviereck (changes requested):
Ah nice! It's really looking good, I do have some more remarks here and there:)

Week3/More JavaScript.js
FarukTercan: Sorry, I just installed the Prettier extension :)

Week3/More JavaScript.js
FarukTercan: it will be undefined result because there are just 2 category defined. 
So I changed the "category > 3" to "category > 2".

Week3/More JavaScript.js
FarukTercan: Right, it is totally a more practical way :)

Week3/More JavaScript.js
FarukTercan: 👍

Week3/More JavaScript.js
FarukTercan: I redesigned the code using the nested objects.

Week3/More JavaScript.js
marcianoviereck: 👍

marcianoviereck (approved):
Great! The code is looking goood! Nice changes and keep it up!:) 
The homework is approved!

PR #153, opened 2019-08-15
Js1 week3

malnaqeeb:


Week3/1_String.js
```diff
@@ -0,0 +1,13 @@
+
+/*1.1 Add the string to your file and log it.*/
+let myString = "hello,this,is,a,difficult,to,read,sentence";
+console.log(myString);
+
+/*1.2 Log the length of myString.*/
+console.log(myString.length);
+
+/*1.3 The commas make that the sentence is quite hard to read. Find a way to remove the commas from the string and replace them with spaces.*/
+let newStr = myString.replace(/,/g, " ");
```
marcianoviereck: nice! Good solution, but do you think it's better to use `let` or `const`?

Week3/2_Arrays.js
```diff
@@ -0,0 +1,35 @@
...
+
+/*2.7 Jason does not like 'giraffe', delete this animal from the array.*/
+favoriteAnimals.splice(3,1);
+
+/*2.8 Again log your new array.*/
+console.log(favoriteAnimals);
+
+/*2.9 Now if unlike Jim, you don't like 'meerkat' and you want to delete it from the array, but you don't know the position or the index of the item in the array, how can you find it?*/
+function finding_index(element){
+    return element === "meerkat"; //declare function for testing
```
marcianoviereck, line 28: aah, interesting! There is an easier solution for this:). Take a look at [indexOf](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf)

Week3/more_javascript.js
```diff
@@ -0,0 +1,183 @@
+/*1- Create a function that takes 3 arguments and returns the sum of the these arguments.*/
+function sum(x, y ,c) {
+   return x+y+c;
+}
+console.log(sum(2,3,5)); 
+
+/* 2- Create a function named colorCar that receives a color, and prints out, 'a red car' for example.*/
+function colorCar(color){
+console.log("a %o car", color)
```
marcianoviereck: ah, neat! Please don't forget to add `;` at the end of the `console.log`. That will help you avoid small mistakes in the future

Week3/more_javascript.js
```diff
@@ -0,0 +1,183 @@
...
+}
+colorCar("red");
+
+/* 3- Create an object and a function that takes the object as a parameter and prints out all of its properties and values.*/
+
+let myDetails = {
```
marcianoviereck: Do you think it's better to use `const` or `let` here?

Week3/more_javascript.js
```diff
@@ -0,0 +1,183 @@
...
+   Name : "Mosleh",
+   Age : 30,
+   Hoppies: ["bascketball" , "movies"]
+}
+function aboutMe(myDetails){
+return console.log(myDetails);
```
marcianoviereck: You could add `identation` here so the code becomes more readable:

Like this:
```
function aboutMe(myDetails){
    return console.log(myDetails);
}
```

Week3/more_javascript.js
```diff
@@ -0,0 +1,183 @@
...
+}
+aboutMe(myDetails);
+// I tried also to get this result using for loop 
+// for (let key in myDetails){
+//    console.log(key + ": " +myDetails[key]);
```
marcianoviereck, line 26: nice! Both is fine!

Week3/more_javascript.js
```diff
@@ -0,0 +1,183 @@
...
+
+/* 4-Create a function named vehicleType that receives a color, and a code, 1 for car, 2 for motorbike. And prints 'a blue motorbike' for example when called as vehicleType("blue", 2)*/
+
+function vehicleType(color, code){
+  if (code === 1){
+     console.log("a %o car", color )
+   
+   }else if (code ===2)
+   console.log("a %o motorbike", color)
+}
```
marcianoviereck, line 37: Try to keep the `identation` consistent. This will make it easier (for yourself) to read the code. And especially for others:

```
  if (code === 1){
     console.log("a %o car", color );
  } else if (code ===2) {
     console.log("a %o motorbike", color);
  }
}
```

Week3/more_javascript.js
```diff
@@ -0,0 +1,183 @@
...
+   vehicleType("red",2);
+
+/* 5- Can you write the following without the if statement, but with just as a single line with console.log(...);?
+if (3 === 3) {
+  console.log("yes");
+} else {
+  console.log("no");
+}*/
+
+switch (3===3){
```
marcianoviereck, line 47: Interesting, but this is not in a `single line`. Can you take a look at this again?

Week3/more_javascript.js
```diff
@@ -0,0 +1,183 @@
...
+// I tried also to get this result using for loop 
+// for (let key in myDetails){
+//    console.log(key + ": " +myDetails[key]);
+// }
+
+/* 4-Create a function named vehicleType that receives a color, and a code, 1 for car, 2 for motorbike. And prints 'a blue motorbike' for example when called as vehicleType("blue", 2)*/
+
+function vehicleType(color, code){
+  if (code === 1){
+     console.log("a %o car", color )
```
marcianoviereck, line 33: I think that `%s` also works fine, just a suggestion.

Week3/more_javascript.js
```diff
@@ -0,0 +1,183 @@
...
+function vehicle(color, code, age){
+   if (code === 1 && age > 0 ){
+      console.log("a %o used car", color)
+  } else if (code ===2 && age > 0)
+    console.log("a %o used motorbike", color)
+    else if (code ===1 && age === 0)
+    console.log("a %o new motorbike", color)
+    else if (code ===2 && age === 0)
+    console.log("a %o new motorbike", color)
+}
```
marcianoviereck, line 67: Here the same comment as before (with the `identation`). And can you also use `{` and `}` around your `else if` statements. Because it's easy to make little mistakes in this way.

And I think that this assignment is supposed to use the `myVehicles` array. So can you take a look at this again?

Week3/more_javascript.js
```diff
@@ -0,0 +1,183 @@
...
+
+console.log(myVehicles[2]); // third element has index 2
+
+/* 9- Change the function vehicle to use the list of question 7. So thxat vehicle("green", 3, 1) prints "a green new bike".*/
+
+let i=0;
+function vehicles(color,age,i){
+  let index = myVehicles[i];
+  if (age > 0 ){
+      console.log("a %o used %o",color,index)
```
marcianoviereck: I don't think you have to use the `index` here, right? You can also directly do this:
`console.log("a %o used %o", color, myVehicles[i]);`?

Week3/more_javascript.js
```diff
@@ -0,0 +1,183 @@
...
+  }   
+   else if ( age === 0 )
+      console.log ("a %o new %o",color,index )
+}
+vehicles("green",0,2);
+
+/* 10- Use the list of vehicles to write an advertisement. So that it prints something like: "Amazing Joe's Garage, we service cars, motorbikes, caravans and bikes."*/
+function advertisement (arr) {
+   let last = arr.slice(-1);
+   return arr.slice(0, arr.length - 1).join('s, ') + 's & ' + last+'s';
```
marcianoviereck: wow! super nice solution:)

Week3/more_javascript.js
```diff
@@ -0,0 +1,183 @@
...
+},{
+   Name: 'Unmes',
+   language: 'CLI',
+},{
+   Name: 'Yash',
+   language: 'Javascript1',
+}]
+console.log(teachers);
+
+/*15- Test the following:
```
marcianoviereck: Why have you commented out this part of the assignment?

Week3/more_javascript.js
```diff
@@ -0,0 +1,183 @@
...
+myVehicles.push("mopid");
+console.log("Amazing Joe's Garage, we service "+advertisement(myVehicles));
+
+/*12- Create an empty object.*/
+let teachers1 ={};
+let teachers2 = new Object();
+console.log(teachers1);
+console.log(teachers2);
+/* 13- Create an object that contains the teachers that you have had so far for the different modules.*/ 
+let teachers=[{
```
marcianoviereck: This looks nice! But now you've created an array with `objects` inside of it, but I think the assignment is to create an `object` and to have an array inside of it.

marcianoviereck (changes requested):
The homework is looking really good:) keep it up! I do have some feedback here and there:)

Week3/2_Arrays.js
malnaqeeb: you have right, this seems much easier :)

`console.log(favoriteAnimals.indexOf('meerkat'));`

Week3/more_javascript.js
malnaqeeb: done :+1:

Week3/more_javascript.js
malnaqeeb: const of course :)

Week3/more_javascript.js
malnaqeeb: done :+1:

Week3/more_javascript.js
malnaqeeb: Thanks

Week3/more_javascript.js
malnaqeeb: oops!! i did'nt read it good 
I think you mean in this way : `console.log((3===3 ? 'Yes' :'No'));`

Week3/more_javascript.js
malnaqeeb: intresting

Week3/more_javascript.js
marcianoviereck: Yep:)

If you can change this in your homework code and address my other comments, then I will re-check the homework!

Week3/more_javascript.js
malnaqeeb: I think that we have to use `myVehicles` array in step 7, is it right?

Week3/more_javascript.js
malnaqeeb: yes, I don't what i was thinking when i did that :)

Week3/more_javascript.js
```diff
@@ -80,19 +80,19 @@ console.log(myVehicles[2]); // third element has index 2
...
 }
 vehicles("green",0,2);
 
 /* 10- Use the list of vehicles to write an advertisement. So that it prints something like: "Amazing Joe's Garage, we service cars, motorbikes, caravans and bikes."*/
 function advertisement (arr) {
-   let last = arr.slice(-1);
-   return arr.slice(0, arr.length - 1).join('s, ') + 's & ' + last+'s';
+   let first =  myVehicles.slice(0, myVehicles.length - 1);
+   let last = myVehicles.slice(-1);
+   return first.join('s, ') + 's & ' + last+'s';
```
marcianoviereck, line 95: Nice!

Week3/more_javascript.js
```diff
@@ -101,36 +101,22 @@ myVehicles.push("mopid");
...
-   language: 'HTML/CSS',
-},{
-   Name: 'Unmes',
-   language: 'CLI',
-},{
-   Name: 'Yash',
-   language: 'Javascript1',
-}]
-console.log(teachers);
+let teachers1={
```
marcianoviereck, line 125: Ah, actually the old version of the code was correct! I misunderstood the assignment before, but after talking to Yash, I realized that the old way of your code is correct!

```
teachers=[{
   Name: 'Rob',
   language: 'HTML/CSS',
},{
   Name: 'Philipp',
   language: 'HTML/CSS',
},{
   Name: 'Unmes',
   language: 'CLI',
},{
   Name: 'Yash',
   language: 'Javascript1',
}]
console.log(teachers);
```

Week3/more_javascript.js
marcianoviereck: I don't think you've addressed this question:)

marcianoviereck (changes requested):
The homework is looking really good:) I'm only not sure why you've commented out question 15 and 16. I understand that you explain the solution in the comments, but can you also show it in the code?

Week3/more_javascript.js
```diff
@@ -106,27 +106,41 @@ let secondMethod = new Object();
...
-z == x*/ 
+// What do you think will happen with
+// x == y
+console.log(x == y); // false  
+// x === y
+console.log(x === y); //false
+// z==y
+console.log(z == y);  // true
+// z==x
+console.log(z == x); // false
```
marcianoviereck, line 143: Ah nice! Could you also uncomment assignment `16` and show the code for your explanation?

marcianoviereck (changes requested):
Awesome that you fixed the merge conflict:) Could you also uncomment assignment 16 and show your code for the explanation?

Week3/more_javascript.js
malnaqeeb: Hi @marcianoviereck 
Thanks for your feedback. I have uncommented this part, but actually this is not my code, it's part of the question :)

marcianoviereck (approved):
Nice! The homework is approved :) keep it up!

PR #154, opened 2019-08-15
JS1/Week3/Lama

ghost:


Week3/app.js
```diff
@@ -0,0 +1,104 @@
...
+        name: 'Angelique Lyall'
+    }, {
+        name: 'Stacie Dyar'
+    }];
+
+
+    people.sort(function (a, b) {
+        return a.name.localeCompare(b.name);
+    });
+    console.log(people);
```
yash-kapila: 👍

Week3/app.js
```diff
@@ -0,0 +1,104 @@
...
+     */
+
+    const myArray = [13, 1, 4, 6, 8, 3, 2, 2];
+    let evenArray = [];
+    for (let i = 0; i < myArray.length; i++) {
+        if (myArray[i] % 2 === 0) {
+            evenArray.push(myArray[i]);
+        }
+    }
+    console.log(evenArray);
```
yash-kapila: 1. Ah. The logic is correct but you missed the part where it is asked to _create a function_.
2. You can also use `forEach` in this case as we need to iterate over the entire array. `forEach` isn't useful if we need to break out of a loop based on some condition in which case a for loop or for-of loop makes more sense. 🙂

Week3/app.js
```diff
@@ -0,0 +1,104 @@
...
+
+
+    function average() {
+        let sum = 0;
+        for (i = 0; i < students.length; i++) {
+            sum += students[i].score;
+        }
+        const scoreAverage = (sum / students.length).toFixed(2);
+        return scoreAverage;
+    }
```
yash-kapila: 1. Would it not be better to pass `students` as an argument to the function instead of using it directly inside `average` function. Because currently, I won't be able to reuse your function with a different array.
2. Also, try to create function names with a verb at the beginning describing what action is being performed inside it. Something like:

```
function calculateAverage(input) {
  ... repeat logic but using `input` instead of `students`
}

console.log(calculateAverage(students));
```

Is it clear? 🙂

Week3/app.js
```diff
@@ -0,0 +1,104 @@
...
+
+
+    /* Q4.
+    * Write a JavaScript function which accept a string as argument and swaps the case of each character.
+    * For example if you input 'The Quick Brown Fox' the output should be 'tHE qUICK bROWN fOX'.
+    * Any character other than alphabets will stay as it is.
+    * Hint: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charCodeAt
+    */
+    function convert(str) {
+        let arr = str.split('');
```
yash-kapila: This is fine but you can also avoid creating a new variable here to split the entire string to an array. You can also perform this operation inside the `for` loop such as:
```
for (let i = 0; i < str.split('').length; i++) {
    ... loop body
}
```

Week3/app.js
```diff
@@ -0,0 +1,104 @@
...
+        let output = [];
+        let temp;
+        for (let i = 0; i < arr.length; i++) {
+            if (arr[i].charCodeAt(0) >= 97 && arr[i].charCodeAt(0) <= 122) {
+                temp = String.fromCharCode((arr[i].charCodeAt(0) - 32))
```
yash-kapila: How about using methods `toUpperCase()` and `toLowerCase()` here instead. I think it would be easier to understand. What do you think? 🙂

Week3/morejava.js
```diff
@@ -0,0 +1,181 @@
...
+console.log(sum(1, 2, 3));
+
+//Create a function named colorCar that receives a color, and prints out, 'a red car' for example.
+function colorCar(x) {
+    return 'a ' + x + ' car';
+}
+console.log(colorCar('red'))
+
+//Create an object and a function that takes the object as a parameter and prints out all of its properties and values.
+let funObject = {
```
yash-kapila: I don't see you re-assigning this variable anywhere else so we can also use `const` instead of `let` to create this variable

Week3/morejava.js
```diff
@@ -0,0 +1,181 @@
...
+//And prints 'a blue motorbike' for example when called as vehicleType("blue", 2)
+function vehicleType(color, code) {
+    if (code === 1) {
+        console.log('a ' + color + ' car');
+    }
+    if (code === 2) {
+        console.log('a ' + color + ' motorbike');
+    }
+}
+vehicleType("blue", 2)
```
yash-kapila, line 36: 👍

Week3/morejava.js
```diff
@@ -0,0 +1,181 @@
...
+//Can you write the following without the if statement,
+//but with just as a single line with console.log(...);?
+
+if (3 === 3) {
+    console.log("yes");
+} else {
+    console.log("no");
+}
+
+console.log((3 === 3) ? 'yes' : 'no');
```
yash-kapila: 👍 but make sure that you use conditional operator (`? :`) only if you wish to use the value being returned and not a statement inside i.e

This is understandable:
```
const answer = (3 === 3) ? 'yes' : 'no'
```

This isn't
```
(3 === 3) ? console.log('yes') : console.log('no');
```

Week3/morejava.js
```diff
@@ -0,0 +1,181 @@
...
+
+//Create a function called vehicle, like before, 
+//but takes another parameter called age, so that vehicle("blue", 1, 5) prints 'a blue used car'
+
+function vehicle(color, code, years) {
+    if (code === 1 && years < 2) {
+        console.log('a ' + color + ' new car')
+
```
yash-kapila: Why so many spaces here in between? It doesn't look nice 🙁

Week3/morejava.js
```diff
@@ -0,0 +1,181 @@
...
+        console.log('a ' + color + ' used car')
+
+
+    }
+    if (code == 2 && years < 2) {
+        console.log('a ' + color + ' new motorbike')
+
+
+    }
+    if (code == 2 && years >= 2) {
```
yash-kapila: `===` instead of `==` please 🙂

Week3/morejava.js
```diff
@@ -0,0 +1,181 @@
...
+
+
+    }
+
+}
+vehicle("blue", 1, 5);
+
+//Make a list of vehicles, you can add "motorbike", "caravan", "bike", or more.
+
+listOfVehicles = ["motorbike", "caravan", "bike", "car"];
```
yash-kapila: I can't see a keyword being used while creating `listOfVehicles`. This means it will be created in the global scope which is not what we want, right? 😢

Week3/morejava.js
```diff
@@ -0,0 +1,181 @@
...
+//So that vehicle("green", 3, 1) prints "a green new bike".
+function newVehicle(color, code, years) {
+    if (years < 2) {
+        console.log('a ' + color + ' new ' + listOfVehicles[code - 1]);
+    }
+    if (years >= 2) {
+        console.log('a ' + color + ' used ' + listOfVehicles[code - 1]);
+    }
+  
+}
```
yash-kapila, line 90: 👍

Week3/morejava.js
```diff
@@ -0,0 +1,181 @@
...
+
+//Use the list of vehicles to write an advertisement.
+//So that it prints something like: "Amazing Joe's Garage, we service cars, motorbikes, caravans and bikes.". (Hint: use a for loop.)
+function advertisement(listOfVehicles) {
+    let adv = "Amazing Joe's Garage, we service "
+
+    for (let i = 0; i < listOfVehicles.length; i++) {
+        adv += (listOfVehicles[i] + "s, ")
+    }
+    console.log(adv + listOfVehicles[listOfVehicles.length - 2] + "s and " + listOfVehicles[listOfVehicles.length - 1] + "s.")
```
yash-kapila: Ah. There seems to be a problem in this logic. I see the output as `Amazing Joe's Garage, we service motorbikes, caravans, bikes, cars, bikes and cars.` i.e bikes and cars being printed twice. Can you find out why? I can suggest that you run the loop a bit differently perhaps 😉

Week3/morejava.js
```diff
@@ -0,0 +1,181 @@
...
+const teachersInfo = {
+    names: ["Philip", "Unmesh", "Yash"]
+};
+
+
+console.log(teachersInfo);
+
+//Add a property to the object you just created that contains the languages that they have taught you.
+teachersInfo.languages = ['HTML/CSS', 'CLI', 'JavaScript1'];
+console.log(teachersInfo);
```
yash-kapila, line 121: I find a slight problem here. Your **teachersInfo** object looks like this:
```
const teachersInfo = {
  names: ["Philip", "Unmesh", "Yash"],
  languages: ['HTML/CSS', 'CLI', 'JavaScript1']
};
```

1. You are using both double quotes(") and single quotes(') for creating strings. It's consistent if you pick one and use that everywhere. 🙂 
2. It would be hard to identify a person not knowing HYF to identify which language was taught by which teacher? For example, did Phillip teach JS1 or Unmesh HTML/CSS? 😆 

It's important that we create meaningful data structures to represent information we want to create. Wouldn't the following be a better approach to this?

```
const teachersInfo = {
  Phillip: 'HTML/CSS',
  Unmesh: 'CLI',
  Yash: 'JavaScript1',
};
```

Week3/morejava.js
```diff
@@ -0,0 +1,181 @@
...
+
+
+//Write some code to test two arrays for equality using == and ===. Test the following:
+let x = [1, 2, 3];
+let y = [1, 2, 3];
+let z = y;
+console.log(x == y); //false 
+console.log(x === y); //false 
+console.log(z == y); //true
+console.log(z == x) //false
```
yash-kapila, line 131: 👍

Week3/morejava.js
```diff
@@ -0,0 +1,181 @@
...
+console.log(o2);//foo: "bar"
+console.log(o3);//foo: "bar"
+
+
+//Show that changing o2 changes o3 (or not)
+o2 = {
+    foo: "changed"
+};
+console.log(o2);//foo: "changed"
+console.log(o3);//foo: "bar"
```
yash-kapila, line 149: 👍

Week3/morejava.js
```diff
@@ -0,0 +1,181 @@
...
+
+/*So when I change an object that does not affect any of the others*/
+
+//What does the following code return? (And why?)
+
+let bar = 42;
+typeof typeof bar;
+let theType = typeof bar;
+console.log(theType); //number
+console.log(typeof theType); //string
```
yash-kapila, line 167: 👍

yash-kapila (changes requested):
Hi @lamaajaj - apologies for leaving it late but I've now finished reviewing your homework. I have some comments regarding a few questions and would appreciate if you had a look into them.

Also, I see that you submitted your homework from the `master` branch of your forked repo instead of creating a new branch for your week3's homework submission. This homework submission also contains your homework from Week2 which seems incorrect. Please take care of this in the future. It's very important that you create different branches while submitting your homework for different week's. I don't think teachers in future modules would be able to approve your homework if they find you haven't created branches. Have a good day ahead. 🙂

Week3/app.js
ghost: I have created a function 👍

Week3/morejava.js
ghost: 👍

Week3/morejava.js
ghost: you are right!

Week3/morejava.js
ghost: It is clean now, I think:)

Week3/morejava.js
ghost: 👍

Week3/morejava.js
ghost: yes, absolutely right!

Week3/morejava.js
ghost: Thanks for the hint, I have changed the loop :)

Week3/morejava.js
ghost: Done! 👍

Week3/app.js
ghost: Yes, it is absolutely clear and better this way!Thanks

Week3/app.js
ghost: you are right!! I wrote a new easy function :)

yash-kapila (approved):

PR #159, opened 2019-08-26
submit week3 hw solutions- Abdullah

abdullahkanatsiz7:


Week3/app.js
```diff
@@ -0,0 +1,239 @@
...
+  }
+  console.log(newstring('The Quick Brown Fox'));
+}
+
+//More JavaScript
+
+// 1. Create a function that takes 3 arguments and returns the sum of the these arguments.
+
+function sum(arr) {
+  return arr.reduce(function(a, b, c) {
```
marcianoviereck, line 106: wow, advanced!

Week3/app.js
```diff
@@ -0,0 +1,239 @@
...
+
+console.log(sum(numbers));
+
+// 2. Create a function named colorCar that receives a color, and prints out, 'a red car' for example.
+const colorCar = color => `a ${color} car`;
+console.log(colorCar('red'));
+
+// 3. Create an object and a function that takes the object as a parameter and prints out all of its properties and values.
+
+const carproperties = {
```
marcianoviereck, line 121: I would advise to use `camelCase` here.

So instead of `carproperties ` I would do: `carProperties `

Week3/app.js
```diff
@@ -0,0 +1,239 @@
...
+const vehicles = ['motorbike', 'caravan', 'bike', 'car'];
+
+//8. How do you get the third element from that list?
+
+console.log(vehicles[2]);
+
+//9. Change the function vehicle to use the list of question 7. So that vehicle("green", 3, 1) prints "a green new bike".
+
+let vehic = (color, code, age) => {
+  age === 1 || age === 0 ? (age = 'new') : (age = 'used');
```
marcianoviereck, line 179: Ah, this works yeah! But re-assigning variables that are passed through the function is not recommended:) just so you know

Week3/app.js
```diff
@@ -0,0 +1,239 @@
...
+  const type = vehicles[code];
+  console.log(`a ${color} ${age} ${type}`);
+};
+
+vehic('green', 2, 1);
```
marcianoviereck: What if I would do `vehic('green', 5, 1);` ?

Week3/app.js
```diff
@@ -0,0 +1,239 @@
...
+
+// 10. Use the list of vehicles to write an advertisement. So that it prints something like: "Amazing Joe's Garage, we service cars, motorbikes, caravans and bikes."
+
+const vehicless = ['motorbike', 'car', 'caravan', 'bike'];
+
+newprint = () => {
+  const first = vehicless[3];
```
marcianoviereck: Why is this variable called `first`? because it's the fourth element in the array

Week3/app.js
```diff
@@ -0,0 +1,239 @@
...
+
+  const second = vehicless
+    .map(vehicless => vehicless + 's')
+    .filter((add, index) => add !== first + 's')
+    .join(', ');
+
+  console.log(`Amazing Joe's Garage, we service ${second}  and ${first}s.`);
```
marcianoviereck, line 203: Nice solution!

Week3/app.js
```diff
@@ -0,0 +1,239 @@
...
+};
+
+newprint();
+
+// 11. What if you add one more vehicle to the list, can you have that added to the advertisement without changing the code for question 10?
+
+vehicless.push('boat');
+console.log(vehicless);
```
marcianoviereck: I think you meant to call `newprint();` here again? So that you could see the result?

Week3/app.js
marcianoviereck: And what happens if there are only two vehicles in the list? Does this code still work?

Week3/app.js
```diff
@@ -0,0 +1,239 @@
...
+    courses: 'CLI, Git',
+  },
+  {
+    name: 'Yash Kapila',
+    courses: 'Git, GitHub, JavaScript',
+  },
+];
+console.log(teacherlesson);
+
+// 17.What does the following code return? (And why?)
```
marcianoviereck, line 239: Did you forget assignment `15` and `16`?

Week3/app.js
```diff
@@ -0,0 +1,239 @@
...
+    },
+    {
+      name: 'Stacie Dyar',
+    },
+  ];
+
+  people.sort(function(a, b) {
+    return a.name.localeCompare(b.name);
+  });
+  console.log(people);
```
marcianoviereck, line 31: nice! I'm not sure if this was part of the homework, but because you submitted it I will check it :)

Week3/app.js
```diff
@@ -0,0 +1,239 @@
...
+        secondstring = secondstring.concat(str[i].toUpperCase());
+      } else if (str.charCodeAt(i) >= 65 && str.charCodeAt(i) <= 90) {
+        secondstring = secondstring.concat(str[i].toLowerCase());
+      } else {
+        secondstring = secondstring.concat(str[i]);
+      }
+    }
+    return secondstring;
+  }
+  console.log(newstring('The Quick Brown Fox'));
```
marcianoviereck, line 98: Aha, interesting! And the code is looking really good :)

marcianoviereck (changes requested):
The homework is looking really good and I see that you already know many advanced Javascript, awesome! Keep it up. I only have some small remarks here and there and I think that you forgot assignment 15 and 16

Week3/app.js
abdullahkanatsiz7: yes it must be like that 
vehicless.push('boat');
console.log(newPrint());

Week3/app.js
abdullahkanatsiz7: Actually, I didn't forget assignments 15 and 16. I had sent both of previously homework that's why  I didn't send again but if you want I could be sent

Week3/app.js
abdullahkanatsiz7: this part from Yash. He asks to us make solution this question

Week3/app.js
abdullahkanatsiz7: yeah, I will use your recommended next time

Week3/app.js
abdullahkanatsiz7: the result is: a green new undefined because I took all strings from vehic array .there are 5 string in this array when I call fifth string I didn't reach anything.what should the answer be ?

Week3/app.js
abdullahkanatsiz7: yes the correct result is
 const first = vehicless[vehicless.length - 1];

Week3/app.js
abdullahkanatsiz7: the result is: a green new undefined because I took all strings from vehic array .there are 5 string in this array when I call fifth string I didn't reach anything.what should the answer be ?

Week3/app.js
marcianoviereck: Ah okay:) then it's fine!

Week3/app.js
marcianoviereck: yeah, could you update your code with it :)?

Week3/app.js
marcianoviereck: Yeah that's what I mean :). If you add some extra `if-statements` in your function, you can check if the `code` is between 0 and `vehicles.length` if it is not, you can print an error. This is to avoid accidental mistakes

abdullahkanatsiz7 (commented):
this is new update

Week3/app.js
abdullahkanatsiz7: I updated my homework.

Week3/app.js
```diff
@@ -0,0 +1,240 @@
...
+//8. How do you get the third element from that list?
+
+console.log(vehicles[2]);
+
+//9. Change the function vehicle to use the list of question 7. So that vehicle("green", 3, 1) prints "a green new bike".
+
+let vehic = (color, code, age) => {
+  age === 1 || age === 0 ? (age = 'new') : (age = 'used');
+  const type = vehicles[code];
+  type.length ? vehicles[code] : 'error';
```
marcianoviereck: Ah, did you mean this:

```
code < type.length ? vehicles[code] : 'error' ?

Because now you check if the type has a `length` and if it doesn't have one, then you show 'error'. It's almost correct, but not quite yet
```

marcianoviereck (changes requested):
One small request left! Then It's approved:)

Week3/app.js
abdullahkanatsiz7: Yes, there is a mistake. Now I changed my function
let vehic = (color, code, age) => {
  age === 1 || age === 0 ? (age = 'new') : (age = 'used');
  const type = vehicles[code];
  if(code>3 || code<0){
    console.log('error');
  }
  else{ 
  console.log(`a ${color} ${age} ${type}`);
}
};

vehic('green', 3, 1);

marcianoviereck (approved):
The homework is looking really nice! keep it up!

abdullahkanatsiz7:
Thank you for all this comment.You spend many time for check and reviewed our homework and thanks to this we going to improve ourselves. Thank you again.

PR #161, opened 2019-09-02
JS1-Week3-HomeWork

husenjanobul:


Week3/app.js
```diff
@@ -0,0 +1,144 @@
+// 1.Strings
+let myString = "hello,this,is,a,difficult,to,read,sentence";
+console.log(myString);
+console.log(myString.length);
+const newStr = myString.split(",");
+console.log(newStr);
+const strStr = newStr.join(" ");
+console.log(strStr);
```
yash-kapila, line 8: Hi @husenjanobul - couple things here.
1. We can also use `const` keyword instead of `let` for `myString` variable like you have done for `newStr` and `strStr`, isn't it?
2. Suggestion: We can also combine these operations in a single one like: `const newString = myString.split(',').join(' ');`. This has a benefit that we don't need to create two new variables `newStr` and `strStr` for it.

Week3/app.js
```diff
@@ -0,0 +1,144 @@
...
+
+// 2.Arrays
+let favoriteAnimals = ["blowfish", "capricorn", "giraffe"];
```
yash-kapila: `const` instead of `let` should also work fine here because we aren't reassigning `favoriteAnimals` variable 🙂

Week3/app.js
```diff
@@ -0,0 +1,144 @@
...
+favoriteAnimals.push("turtle");
+console.log(favoriteAnimals);
+favoriteAnimals.splice(1, 0, "meerkat");
+// new arry  ["blowfish", "meerkat", "capricorn", "giraffe","turtle"]
+console.log(favoriteAnimals);
+console.log("The array has a length of: " + favoriteAnimals.length);
+favoriteAnimals.splice(3, 1);
+console.log(favoriteAnimals);
+// 2.9 ~2.10
+for (let a = 0; a < favoriteAnimals.length; a++) {
```
yash-kapila, line 22: You can also explore `indexOf`(https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf) or `find`(https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) methods for this operation 🙂

Week3/app.js
```diff
@@ -0,0 +1,144 @@
...
+        console.log("The item you are looking for is at index: " + a);
+    }
+}
+
+// More JavaScript 
+// 1.Create a function that takes 3 arguments and returns the sum of the these arguments.
+function sumArg(arg1, arg2, arg3) {
+    // let sum = arg1 + arg2 + arg3;
+    // return sum;
+    return arg1 + arg2 + arg3;
```
yash-kapila, line 33: 👍 but would suggest exploring default parameters of ES6 as well - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

Week3/app.js
```diff
@@ -0,0 +1,144 @@
...
+}
+console.log(sumArg(1, 2, 3));
+
+// 2.Create a function named colorCar that receives a color, and prints out, 'a red car' for example.
+function colorCar(color) {
+    let carColor = "a " + color + " car";
+    return carColor;
+}
+console.log(colorCar("red"));
```
yash-kapila, line 42: 👍 but try to make use of `const` instead of `let` please for variables which aren't going to be reassigned.

Week3/app.js
```diff
@@ -0,0 +1,144 @@
...
+
+// 3.Create an object and a function that takes the object as a parameter and prints out all of its properties and values.
+const testObj = {
+    s: "Name",
+    d: "Age",
+    f: "DOB"
+};
+function proObj(obj) {
+    for (let [key, value] of Object.entries(obj)) {
```
yash-kapila, line 51: 1. `const` here instead of `let` would also work fine here.
2. The function `proObj` isn't the best. We should choose function names in such a way that they explain what logic they contain without having to read their body. In this case, I had to read the code to understand that you are simply printing the key-value pair of the argument. Perhaps a better name would have been `printKeyValues` or `printObjectProps`.

Week3/app.js
```diff
@@ -0,0 +1,144 @@
...
+        console.log(`${key}: ${value}`);
+    }
+}
+proObj(testObj);
+
+// 4.Create a function named vehicleType that receives a color, and a code, 1 for car, 2 for motorbike. And prints 'a blue motorbike' for example when called as vehicleType("blue", 2)
+let codeVehicle = {
```
yash-kapila, line 58: `const` here please

Week3/app.js
```diff
@@ -0,0 +1,144 @@
...
+    1: "car",
+    2: 'motorbike'
+};
+function vehicleType(color, code) {
+    console.log("a " + color + " " + codeVehicle[code]);
+}
+vehicleType("blue", 2);
+
+// 5.Can you write the following without the if statement, but with just as a single line with console.log(...);?
+let h = (3 === 3) ? "yes" : "no";
```
yash-kapila, line 68: `const` here please

Week3/app.js
```diff
@@ -0,0 +1,144 @@
...
+console.log(h);
+
+// 6.Create a function called vehicle, like before, but takes another parameter called age, so that vehicle("blue", 1, 5) prints 'a blue used car'
+function vehicle(color, code, age) {
+    if (age === 1) {
+        console.log("a " + color + " " + "new " + codeVehicle[code]);
+    } else{
```
yash-kapila, line 75: One space missing between `else{` 😉

Week3/app.js
```diff
@@ -0,0 +1,144 @@
...
+        console.log("a " + color + " " + "used " + codeVehicle[code]);
+    }
+}
+vehicle("blue", 1, 5);
+
+// 7.Make a list of vehicles, you can add "motorbike", "caravan", "bike", or more.
+let listVehicles = ["motorbike", "caravan", "bike"];
+// 8.How do you get the third element from that list?
+console.log(listVehicles[2]);
```
yash-kapila, line 84: 👍

Week3/app.js
```diff
@@ -0,0 +1,144 @@
...
+// 10.Use the list of vehicles to write an advertisement. So that it prints something like: "Amazing Joe's Garage, we service cars, motorbikes, caravans and bikes.". (Hint: use a for loop.)
+let advertisement = "Amazing Joe's Garage, we service ";
+for (let q = 0; q < listVehicles.length; q++){
+    if (q < (listVehicles.length-1)){
+        advertisement = advertisement +  ", " + listVehicles[q];
+    } else if (q === (listVehicles.length - 1)){
+        advertisement = advertisement + " and " + listVehicles[q] + ".";
+    }
+}
+console.log(advertisement);
```
yash-kapila, line 105: Hi Huseyin - there seems to be a problem here. The expected output was: `Amazing Joe's Garage, we service cars, motorbikes, caravans and bikes.` but what I see happening is: `Amazing Joe's Garage, we service , motorbike, caravan and bike.`. Could you please check?

Week3/app.js
```diff
@@ -0,0 +1,144 @@
...
+// 13.Create an object that contains the teachers that you have had so far for the different modules.
+// 14.Add a property to the object you just created that contains the languages that they have taught you.
+let myTeachers = {
+    HTML : "Philipp",
+    CSS : "Rob",
+    CLI : "Unmesh",
+    JS1 : "Yash"
+};
+// console.log(myTeachers);
+myTeachers["Git"] = "Yash";
```
yash-kapila, line 123: It was also Bonan for Git, isn't it? 😉

Week3/app.js
```diff
@@ -0,0 +1,144 @@
...
+// console.log(myTeachers);
+
+// 15.Write some code to test two arrays for equality using == and ===. Test the following:
+let x = [1, 2, 3];
+let y = [1, 2, 3];
+let z = y;
+// What do you think will happen with x == y, x === y and z == y and z == x? Prove it!
+// x == y  and x === y  ----  false , x and y are arrays and complex data type, they're store different places in Ram.
+// z == y -------- true , z and y pointed same place.
+// z == x -------- false , they're different place in Ram
```
yash-kapila, line 133: 👍

Week3/app.js
```diff
@@ -0,0 +1,144 @@
...
+
+const o1 = { foo: "bar" };
+const o2 = { foo: "bar" };
+const o3 = o2;
+// i think o2 changes o3 --( o3 = o2 )  left side(o3) declare by right side(o2), and they're same location in the Ram.
+// o1 changes o3
```
yash-kapila: Are you sure `o1` changes `o3`?

Week3/app.js
```diff
@@ -0,0 +1,144 @@
...
+const o1 = { foo: "bar" };
+const o2 = { foo: "bar" };
+const o3 = o2;
+// i think o2 changes o3 --( o3 = o2 )  left side(o3) declare by right side(o2), and they're same location in the Ram.
+// o1 changes o3 
+//but  i think the order is important--( o2 = o3 )  left side(o2) already declared 
+
+// 17.What does the following code return? (And why?)
+let bar = 42;
+typeof typeof bar;
```
yash-kapila, line 144: 👍

yash-kapila (changes requested):
Hi @husenjanobul - I've finished reviewing your homework. There are a couple of questions which I've raised towards the end. Please have a look at it. Other than that, I have reiterated using `const` instead of `let` wherever we aren't re-assigning a variable to a different value. Please take care of it next time onwards. Good luck with JS2-Week2's homework and see you next Sunday 🙂

Week3/app.js
husenjanobul: Thanks your feedback .

Week3/app.js
husenjanobul: okey,  i'll check it and improve.

Week3/app.js
husenjanobul: no , they'er different location in the memory.

Week3/app.js
husenjanobul: i'll set default value.

PR #168, opened 2019-09-20
homework week1 nuri

nbayram:


Week1/homework/js-exercises/logArray.js
```diff
@@ -0,0 +1,11 @@
...
+
+{
+  let myClothes = [];
+  console.log('pants', 'shirt', 'shoes');
+  console.log(myClothes);
+  let myArrAnimal = ['Giraffe', 'Elephant', 'Lion'];
+  console.log(myArrAnimal);
+  myArrAnimal.push('Piglet');
+  console.log(myArrAnimal);
+}
```
yash-kapila, line 22: 👍 for using `use strict` and using curly braces to encapsulate your code. However, I think using `const` instead of `let` for creating those variables is a good idea because we are not reassigning those variables anywhere, right?

Week1/homework/js-exercises/logCompare.js
```diff
@@ -0,0 +1,15 @@
+'use script';
+
+{
+  const myArr1 = ['n', 'u', 'r', 'i'];
+  const myArr2 = [' ', 'b', 'a', 'y', 'r', 'a', 'm'];
+
+  let x = console.log(myArr1.length);
+  let y = console.log(myArr2.length);
+
+  if ((x = y)) {
```
yash-kapila: Aahaan. There is a mistake here. Please have a careful look. 🙂 

Also, we can use `const` instead of `let` while creating `x` and `y` since we needn't re-assign them.

Week1/homework/js-exercises/logError.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+
+{
+  // console.log('I'm awesome');
+  // we have to escape after I or we can use quote in double quote.
+  const strAwe = "I'm awesome!";
+  console.log(strAwe);
```
yash-kapila, line 7: 👍

Week1/homework/js-exercises/logHello.js
```diff
@@ -0,0 +1,33 @@
...
+  console.log(strChi);
+
+  const strJap = 'こんにちは世界!';
+  console.log(strJap);
+
+  const strIta = 'Ciao mondo!';
+  console.log(strIta);
+
+  const strRus = 'Привет, мир!';
+  console.log(strRus);
```
yash-kapila: 👍 but this seems to be a case where you could also use a complex data type like arrays or objects and avoid creating so many variables. Something like:

```
const languages = {
  spa: 'Spanish translation',
  ned: 'Nederlands translation',
  // and so on
};

or

const languages = ['Spanish translation', 'Nederlands translation', ..and so on];
```

Week1/homework/js-exercises/logLength.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
+
+{
+  let mySentence = 'Programming is so interesting!';
```
yash-kapila: `const` here instead of `let` please 🙂

Week1/homework/js-exercises/logNumber.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+{
+  let x;
+  console.log(undefined);
+  console.log(x);
+  x = 13;
+  console.log(13);
+  console.log(x);
```
yash-kapila, line 9: 👍

Week1/homework/js-exercises/logRemainder.js
```diff
@@ -0,0 +1,18 @@
+'use script';
+
+{
+  if ((x = 7)) {
```
yash-kapila: Ah. This logic works but it is not a good idea to do an assignment operation inside an `if` block. Could you please revisit this?

Also, if you look carefully, the assignment says:
```
1. If x equals 7, and the only other statement is x = x % 3, what would be the value of x after the calculation?
2. If y equals 21, and the only other statement is x = x % 4, what would be the value of y after the calculation?
3. If z equals 13, and the only other statement is x = x % 2, what would be the value of z after the calculation?
```

Have a careful look 😉

Week1/homework/js-exercises/logRound.js
```diff
@@ -0,0 +1,15 @@
...
+  console.log(z);
+
+  let a = 7;
+  console.log(a);
+
+  if (z > a) {
+    console.log(z);
+  } else {
+    console.log(a);
+  }
```
yash-kapila, line 29: Can use `const` here instead of `let` as well

Week1/homework/js-exercises/logString.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+{
+  let myString = 'Nurulah Bayram';
+  console.log('Nurullah Bayram');
+  console.log(myString);
+  myString = 'David Ng';
+  console.log('David Ng');
+  console.log(myString);
```
yash-kapila, line 9: 👍

Week1/homework/js-exercises/logType.js
```diff
@@ -0,0 +1,59 @@
+'use strict';
+
+{
+  // declare all variables
+  let strClass = 'there is tables and chairs.';
+  let strComputer = 'Computers have a lot of pieces.';
```
yash-kapila: Why not create these string variables as `const` as well?

yash-kapila (changes requested):
Hi @nbayram - I've finished reviewing your homework. It looks good but there are a couple small questions which I would like you to revisit as I think you can improve there. Could you please have a look at them and let me know through a message on Slack once you have resubmitted your changes?

wouterkleijn:
Thanks for the review @yash-kapila and @nbayram for the changes. Did you ask @yash-kapila for a final look?

nbayram:
Yes @wouterkleijn I wrote to @yash-kapila . If he has a time, he will check again.

PR #178, opened 2019-09-22
Clarify homework week 2 exercise 5

nudge:
Clarify the wording of exercise 5 to make it clearer.


NoerGitKat (approved):
As discussed, this is much clearer. Thanks for the clarification, David!

PR #184, opened 2019-09-29
Homework week 1 aline pereira

alineafpereira:


Week1/homework/exercise1.js
```diff
@@ -0,0 +1,10 @@
+console.log("Hello, world! // English")
+console.log("Oi, mundo! // Portuguese")
+console.log("Hola, mundo! // Spanish")
+console.log("Halo, dunia! // Indonesian")
+console.log("Ciao, mondo! // Italian")
+console.log("Hallo, wereld! // Nederlandse")
+console.log("Merhaba, dunya! // Turkish")
+console.log("Bonjour, le monde! // French")
+console.log("Hej, verden! // Denmark")
+console.log("Geia, kósmo! // Greek")
```
yash-kapila: 👍 but I think you meant to have the comments (`// English`) outside the `console.log`, right?

Week1/homework/exercise10.js
```diff
@@ -0,0 +1,11 @@
...
+const itens2 = [45, 78, true, false, {food: "bring me food"}, 65, true];
+console.log ('The first array has the lenght of' + itens1.length);
+console.log ('The second array has the lenght of' + itens2.length);
+
+if (itens1.length == itens2.length)
+{
+    console.log("The arrays have the same length!")
+}   else {
+    console.log("The arrays don't have the same length!")
+}
```
yash-kapila: Ah. 👍 for using `const` instead of `let` but there seems to be a spelling mistake in naming `items` as `itens`. Also, would recommend that you use triple equals(`===`) operator instead of double equals(`==`) when doing comparisons in JavaScript. 🙂

Week1/homework/exercise2.js
```diff
@@ -0,0 +1 @@
+console.log ("I'm awesome")
```
yash-kapila, line 3: 👍

Week1/homework/exercise3.js
```diff
@@ -0,0 +1,8 @@
+let numberx = []
+console.log ("x is undefined because I didn't atribute any value to it");
+console.log (numberx);
+var numberx= 3;
+console.log ("x is equal to 3");
+console.log (numberx);
```
yash-kapila: Ah. There seems to be a mistake.
- You declared the variable `numberx` and initialised it with an empty array(`[]`). Thus, it is not **undefined** anymore but has a value which is an empty array.
- Once you have declared a variable, you need not re-declare it as you do in line 4. Also, try to avoid using `var` keywords while declaring variables.

Week1/homework/exercise4.js
```diff
@@ -0,0 +1,6 @@
+var myString = ("Aline Amaral Farah Pereira")
+console.log ("The string is my name")
+console.log (myString)
+var myString = ("This is not my name anymore")
```
yash-kapila: Couple things similar to before:
- You should avoid using `var` keyword and use `let` or `const` instead
- You need not re-declare a variable once it has been created. Thus, using `var` on line 4 isn't required.

Week1/homework/exercise5.js
```diff
@@ -0,0 +1,14 @@
+const z = 7.25;
+console.log (z);
+const a = 7;
+console.log (a) ;
+if (z > a)
+    
+{
+        const b = z
+        console.log(b)
+}
```
yash-kapila: 👍 but there seems to be an indentation problem here with some extra spacing and new lines which aren't required. 🙂

Week1/homework/exercise6.js
```diff
@@ -0,0 +1,7 @@
+var items = [];
+console.log ("An array with more than one value");
+console.log (items);
+var favanimals = ["dog", "giraffe", "horse"];
+console.log (favanimals);
+favanimals.push("Piglet");
```
yash-kapila: Couple things here again similar to before:
- `const` or `let` instead of `var` keyword.
- If a variable name contains multiple words such as `favanimals` in this case, it is better to use `camelCasing` while naming them. Thus, `favAnimals` would be a better way to name the variable as compared to `favanimals`.

Week1/homework/exercise9.js
```diff
@@ -0,0 +1,10 @@
+let x = 7;
+let y = 21;
+let z = 13;
+
+console.log (x=x%3)
+// the equation above gives us x as 1, once the rest of 7 divided by 3 is 1 (2x3=6 and rests 1 to 7)
+console.log (y=y%4)
+// the equation above gives us y as 0, once the rest of 21 divided by 4 is 5.25 (4x5.25=21 and there is no rest)
+console.log (z=z%2)
```
yash-kapila, line 11: 👍

Week1/homework/exercise8.js
```diff
@@ -0,0 +1,60 @@
+var item1 = ("On wednesday we wear pink.");
+var item2 = ("Except I am wearing red...");
+var item3 = {HeyHo};
+var item4 = {LetsGo};
```
yash-kapila: `const` here please instead of `var`

Week1/homework/exercise8.js
```diff
@@ -0,0 +1,60 @@
...
+
+if (typeof item1 == typeof item2)
```
yash-kapila: Try to make it a habit of using triple equals instead of double equals when doing comparison in JavaScript. 🙂

yash-kapila (changes requested):
Hi @alineafpereira - I've reviewed your homework but there are a couple things which I think would be nice if you implemented. Could you please have a look and let me know if you have any questions?

alineafpereira:
Hey, @yash-kapila! Thanks for reviewing all this exercises. I guess I fixed all the aspects your pointed out. Nice to realize that, after some classes, this challenges doesn't look so hard anymore. I was glad that I could understand all your considerations :)
Anyways, I hope everything is correct now. 

Question: after commiting and pushing the files with all due corrections, do I need to pull the request again or what I've done here is enough for you to see everything? I am still struggling with git/github.

Week1/homework/exercise10.js
alineafpereira: haha, itens in Portuguese meanings items, I guess that's why I misspelled. Thanks for letting me know!

PR #195, opened 2019-11-26
Abuzer alaca

abuz33:
I finished the HW. Thanks a lot.

### DIFF
--- a/Week1/Homework/js-exercises/compareArrays.js
+++ b/Week1/Homework/js-exercises/compareArrays.js
@@ -1,0 +1,22 @@
+"use strict";
+
+var me = ["Abuzer", 29, "unemployed", "bachelor graduate"];
+
+var tesla = [
+  "model S",
+  "model X",
+  "model Y",
+  "model 3",
+  "Roadster",
+  "Cyber Truck",
+  "Solar Panel"
+];
+
+console.log("length of the array me is ", me.length);
+console.log("length of the array tesla is ", tesla.length);
+
+if (me.length == tesla.length) {
+  console.log("They are the same!");
+} else {
+  console.log("Two different sizes");
+}
--- a/Week1/Homework/js-exercises/compareArrays.js
+++ b/Week1/Homework/js-exercises/compareArrays.js
@@ -1,8 +1,8 @@
 "use strict";
 
-var me = ["Abuzer", 29, "unemployed", "bachelor graduate"];
+let me = ["Abuzer", 29, "unemployed", "bachelor graduate"];
 
-var tesla = [
+let teslaModels = [
   "model S",
   "model X",
   "model Y",
@@ -15,7 +15,7 @@ var tesla = [
 console.log("length of the array me is ", me.length);
 console.log("length of the array tesla is ", tesla.length);
 
-if (me.length == tesla.length) {
+if (me.length === tesla.length) {
   console.log("They are the same!");
 } else {
   console.log("Two different sizes");
--- a/Week1/Homework/js-exercises/logAnimals.js
+++ b/Week1/Homework/js-exercises/logAnimals.js
@@ -1,0 +1,12 @@
+"use strict";
+
+var emptyArrays = [];
+
+console.log("It is an empty array.");
+console.log(emptyArrays);
+
+var animals = ["Dog", "Cat", "Lion"];
+console.log(animals);
+
+animals.push("Piglet");
+console.log(animals);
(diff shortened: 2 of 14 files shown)


Week1/Homework/js-exercises/compareArrays.js
```diff
@@ -0,0 +1,22 @@
+"use strict";
+
+var me = ["Abuzer", 29, "unemployed", "bachelor graduate"];
+
+var tesla = [
```
marcianoviereck: Nice, I think a more descriptive name could be "teslaModels" (especially because of the **s** because an array contains multiple items, so usually the variable is in the plural:))

Week1/Homework/js-exercises/compareArrays.js
```diff
@@ -0,0 +1,22 @@
...
+  "model 3",
+  "Roadster",
+  "Cyber Truck",
+  "Solar Panel"
+];
+
+console.log("length of the array me is ", me.length);
+console.log("length of the array tesla is ", tesla.length);
+
+if (me.length == tesla.length) {
```
marcianoviereck: Ah, I think this is almost correct, in Javascript there is a difference between `==` and `===`  See [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators)

Week1/Homework/js-exercises/logAnimals.js
```diff
@@ -0,0 +1,12 @@
+"use strict";
+
+var emptyArrays = [];
```
marcianoviereck: Ah nice, here you do have the `s`!

Week1/Homework/js-exercises/logNumber.js
```diff
@@ -0,0 +1,15 @@
+"usestrict";
+
+var numberX;
+
+console.log(
+  "The numberX will be undefined because we didn't add any value to numberX."
```
marcianoviereck, line 6: nice!

Week1/Homework/js-exercises/compareArrays.js
```diff
@@ -0,0 +1,22 @@
+"use strict";
+
+var me = ["Abuzer", 29, "unemployed", "bachelor graduate"];
```
marcianoviereck: Could you use `const` or `let` instead of `var`? (if you develop this habbit of not-using var, it will be easier in the future ;))

This is also for the rest of the homework

Week1/Homework/js-exercises/logNumber.js
```diff
@@ -0,0 +1,15 @@
+"usestrict";
+
+var numberX;
```
marcianoviereck: Would you use `let` or `const` here?

Week1/Homework/js-exercises/logString.js
```diff
@@ -0,0 +1,14 @@
...
+var myString = "Abuzer ALACA";
+console.log(
+  "The value of myString will be my full name because I assigned my name to it."
+);
+console.log(myString);
+
+myString = "There is an apple.";
+
+console.log(
+  "The value of myString will be a sentence which is 'There is an apple.' because I assigned my name to it."
```
marcianoviereck: Nice, but I think you forgot to remove ` because I assigned my name to it` from the console.log

marcianoviereck (changes requested):
The homework is looking really good! I do have some remarks here and there:)

Week1/Homework/js-exercises/compareArrays.js
abuz33: Ok I will change the name thanks a lot

Week1/Homework/js-exercises/compareArrays.js
abuz33: Yeah I know. == checks the value  and === checks the value and the type. When I used length attribute or method, the result will be an integer for that reason I wrote ==. But I will change it.

abuz33: Yeah I know. == checks the value  and === checks the value and the type. When I used length attribute or method, the result will be an integer for that reason I wrote ==. But I will change it.

Week1/Homework/js-exercises/logNumber.js
abuz33: I have to use `let` because I change the value of numberX later in the program.

marcianoviereck (approved):
The homework is approved now! Nice fixes. But I think you (accidentally?) added the homework for week 2..

So I will approve this one, but that doesn't mean that the homework of week 2 is approved :)

PR #200, opened 2019-11-28
js exercise by ali

aliulgr13:
It needs to be examined.

### DIFF
--- a/Week1/homework/js-exercises/compareArrays.js
+++ b/Week1/homework/js-exercises/compareArrays.js
@@ -1,10 +1,10 @@
 'use strict'
-const bestDefender = ["Van Dijk", "Ramos", "Söyüncü"];
-const bestForward = ["Messi","Ronaldo","Salah","Neymar","Lewandowski","Suarez","Mbappe"];
-console.log('The length of the array is... ' + bestDefender.length);
-console.log('The length of the array is... ' + bestForward.length);
+const bestDefenders = ["Van Dijk", "Ramos", "Söyüncü"];
+const bestForwards = ["Messi","Ronaldo","Salah","Neymar","Lewandowski","Suarez","Mbappe"];
+console.log('The length of the array is... ' + bestDefenders.length);
+console.log('The length of the array is... ' + bestForwards.length);
 
-if (bestDefender.length === bestForward.length){
+if (bestDefenders.length === bestForwards.length){
     console.log("They are the same!")
 } else {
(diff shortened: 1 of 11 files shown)


aliulgr13:
@wilgert  my homework is ok, do I need to change or add something?

Week1/homework/js-exercises/compareArrays.js
```diff
@@ -0,0 +1,11 @@
+'use strict'
+const bestDefender = ["Van Dijk", "Ramos", "Söyüncü"];
```
marcianoviereck: I would add the `s` at the end of the variable, to show that there are more elements in the variable.

So I would change it to: `bestDefenders`

Week1/homework/js-exercises/logLengtString.js
```diff
@@ -0,0 +1,4 @@
+'use strict'
+let mySentence = "Programming is so interesting!";
+let x = mySentence.length;
```
marcianoviereck: Nice, I would change `x` to a name that is more meaningfull, for example: `mySentenceLength`

Week1/homework/js-exercises/roundNumber.js
```diff
@@ -0,0 +1,7 @@
+'use strict'
+let z = 7.25;
+console.log(z);
+let a = Math.round(z);
+console.log(a);
+let highestValue = Math.max(a, z);
```
marcianoviereck, line 6: ah nice! Nice solution

marcianoviereck (changes requested):
Very nice! I only have some small comments on the homework:) If you can fix that then I will approve it.

Week1/homework/js-exercises/compareArrays.js
aliulgr13: > I would add the `s` at the end of the variable, to show that there are more elements in the variable.
> 
> So I would change it to: `bestDefenders`

yeah it is a quite logical : )

Week1/homework/js-exercises/logLengtString.js
aliulgr13: yeah I understand esp. when we are working on a big project, meaningfull name would be useful, thank you.

Week1/homework/js-exercises/roundNumber.js
aliulgr13: Thank you Marciano

wilgert (approved):

PR #206, opened 2019-11-30
finished project for homework week1 form Jalal!

Jalalalattar:
Review :)

### DIFF
--- a/Week1/homework/js-exercises/10-compareArray.js
+++ b/Week1/homework/js-exercises/10-compareArray.js
@@ -12,7 +12,7 @@ if(footballTeam.length === professionalFootballTeam.length){
 
     console.log('They are the same size!');
 
-}else if(footballTeam.length >> professionalFootballTeam.length){
+}else if(footballTeam.length > professionalFootballTeam.length){
 
     console.log(footballTeam.length + 'is smaller than '+professionalFootballTeam.length);
 
--- a/Week1/homework/js-exercises/10-compareArray.js
+++ b/Week1/homework/js-exercises/10-compareArray.js
@@ -1,0 +1,22 @@
+'use strict';
+
+const footballTeam = ['Goalkeeper', 'Defense player', 'Midfielder', 'Attack player'];
+const professionalFootballTeam = ['Goalkeeper','Sweeper (libero)','Defense player', 'Midfielder','Centre forward', 'Attack player'];
+
+console.log('The length of the first array is ...' + footballTeam.length);
+console.log('The length of the second array is ...' + professionalFootballTeam.length);
+
+
+
+if(footballTeam.length === professionalFootballTeam.length){
+
+    console.log('They are the same size!');
+
+}else if(footballTeam.length >> professionalFootballTeam.length){
+
+    console.log(footballTeam.length + 'is smaller than '+professionalFootballTeam.length);
+
+}else{
+
+    console.log(professionalFootballTeam.length + ' is Bigger than ' + footballTeam.length);
+}
--- a/Week1/homework/js-exercises/3-logNumber.js
+++ b/Week1/homework/js-exercises/3-logNumber.js
@@ -5,12 +5,12 @@
 // but if you used another keyword like let or var 
 // var numberX; let numberX; prints undefined without erroe message
 
-var numberX;
+let numberX;
 
 console.log('Guess what is the value of X ? Number or String');
 console.log(numberX); 
 
-var numberX = 7;
+ numberX = 7;
 
 console.log('If you guess it is a number, you are good at guessing , because the value of X is a number');
 console.log(numberX);
--- a/Week1/homework/js-exercises/3-logNumber.js
+++ b/Week1/homework/js-exercises/3-logNumber.js
@@ -1,0 +1,16 @@
+'use strict';
+
+// if you used CONST for a variables without value , You will get an error message
+// const numberX; ( SyntaxError: Missing initializer in const declaration )
+// but if you used another keyword like let or var 
+// var numberX; let numberX; prints undefined without erroe message
+
+var numberX;
+
+console.log('Guess what is the value of X ? Number or String');
+console.log(numberX); 
+
+var numberX = 7;
+
+console.log('If you guess it is a number, you are good at guessing , because the value of X is a number');
+console.log(numberX);
--- a/Week1/homework/js-exercises/4-logString.js
+++ b/Week1/homework/js-exercises/4-logString.js
@@ -1,0 +1,13 @@
+'use strict';
+
+var myString ='Jalal Al Attar';
+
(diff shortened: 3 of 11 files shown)


Week1/homework/js-exercises/10-compareArray.js
```diff
@@ -0,0 +1,22 @@
...
+console.log('The length of the first array is ...' + footballTeam.length);
+console.log('The length of the second array is ...' + professionalFootballTeam.length);
+
+
+
+if(footballTeam.length === professionalFootballTeam.length){
+
+    console.log('They are the same size!');
+
+}else if(footballTeam.length >> professionalFootballTeam.length){
```
marcianoviereck: Ah, you can also just do this:

`footballTeam.length < professionalFootballTeam.length`

Week1/homework/js-exercises/3-logNumber.js
```diff
@@ -0,0 +1,16 @@
...
+// const numberX; ( SyntaxError: Missing initializer in const declaration )
+// but if you used another keyword like let or var 
+// var numberX; let numberX; prints undefined without erroe message
+
+var numberX;
+
+console.log('Guess what is the value of X ? Number or String');
+console.log(numberX); 
+
+var numberX = 7;
```
marcianoviereck: Ah, I think this will give an error (or it should at least), because you've already defined `numberX. Could you take a look at this again?

Week1/homework/js-exercises/4-logString.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+var myString ='Jalal Al Attar';
```
marcianoviereck: Do you think it's better to use `let` or `var` here :)?

Week1/homework/js-exercises/4-logString.js
```diff
@@ -0,0 +1,13 @@
...
+
+console.log('the value of the string is a name, Guess the name !!' );
+console.log(myString);
+
+
+
+var myString ='David Beckham';
```
marcianoviereck: And here the same as before, I think this will give an `already defined` error. Can you take a look at this again?

Week1/homework/js-exercises/5-roundNumber.js
```diff
@@ -0,0 +1,20 @@
...
+
+if (a < z) {
+    const b = z;
+    console.log('The value of the variable Z is the highest');
+    console.log(b);
+}else{
+    const b = a;
+    console.log('The value of the variable A is the highest');
+    console.log(b);
+};
```
marcianoviereck, line 16: Ah nice! You can also do this in an easier way with [Math.max](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max)

Week1/homework/js-exercises/7-logLengthOfString.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+const mySentence = 'national geographic';
+
+console.log(mySentence.length); // the console will print how many letters are written as anumber
+
+for (var i = 0; i<mySentence.length; i++) // This way the console will print the letters individually
+{
+    console.log(mySentence[i]);
```
marcianoviereck, line 9: haha nice!

Week1/homework/js-exercises/9-logTheRemainder.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+/*
+1. If `x` equals 7, and the only other statement is `x = x % 3`, what would be the value of `x` after the calculation?
+2. If `y` equals 21, and the only other statement is `y = y % 4`, what would be the value of `y` after the calculation?
+3. If `z` equals 13, and the only other statement is `z = z % 2`, what would be the value of `z` after the calculation?
+*/
+
+var x = 7;
```
marcianoviereck: Do you think it's better to use `var` or `let` here?

marcianoviereck (changes requested):
The homework is looking really good! I do have some feedback here and there :) But keep it up!

Week1/homework/js-exercises/10-compareArray.js
Jalalalattar: I know I can do that but I liked the options to be more :)

Week1/homework/js-exercises/3-logNumber.js
Jalalalattar: no, there is no error.
it's work because I used **VAR**

Week1/homework/js-exercises/4-logString.js
Jalalalattar: yes because if I used **CONST** I will show this a beautiful  **SyntaxError**
SyntaxError: Identifier 'myString' has already been declared.

Week1/homework/js-exercises/4-logString.js
Jalalalattar: same answer!

Week1/homework/js-exercises/9-logTheRemainder.js
Jalalalattar: Here I can use const.
 No harm in that

Jalalalattar (commented):
**marcianoviereck added ## needs work review in progress labels 2 hours ago****

I think that's not fair
The homework was perfectly ready
It does not need work.

Week1/homework/js-exercises/3-logNumber.js
marcianoviereck: i know you used var, but Wilgert said in the class we should not use var, but `let` or `const`

wilgert:
> **marcianoviereck added ## needs work review in progress labels 2 hours ago****
> 
> I think that's not fair
> The homework was perfectly ready
> It does not need work.

It's totally fair. In modern JavaScript, which we teach at HackYourFuture, we do not use `var`. Instead we use `const` as much as possible and `let` in certain cases where it's really necessary to re-assign a value to a variable.

Week1/homework/js-exercises/10-compareArray.js
wilgert: `>>` is not the greater than operator it is the bitwise operator, see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators

Jalalalattar: The mistakes I made are only using keywords **var* instead of  **let or const**
It happened because I used to use it automatically. But I'll work on the notes.
Thanks guys.

Jalalalattar:
The mistakes I made are only using keywords **var* instead of  **let or const**
It happened because I used to use it automatically. But I'll work on the notes.
Thanks guys.

marcianoviereck (approved):
Nice! Well done, the homework is looking nice and is approved!

PR #208, opened 2019-11-30
Homework week 1 wesam alkfere

wesam-k:


Week1/homework/2-Error debugging.js
```diff
@@ -0,0 +1,4 @@
+'use strick'
```
wilgert: you made a typo here.

Week1/homework/2-Error debugging.js
```diff
@@ -0,0 +1,4 @@
...
+
+
+console.log("I'm awesome");
```
wilgert: This is one way to solve this issue. But I'm going to give you another challenge. What if we have decided we will only use single quotes for strings...
Can you come up with another solution in that case?

Week1/homework/3-Log the number.js
```diff
@@ -0,0 +1,17 @@
+'use strict'
+
+const numberx;
```
wilgert: Did you try to run this code? I think it will throw an error.

Week1/homework/4-Log the string.js
```diff
@@ -0,0 +1,19 @@
+'use strict'
```
wilgert, line 1: Did you try to run this code? I think it will throw an error.

Week1/homework/5-Round a number and log it.js
```diff
@@ -0,0 +1,17 @@
+'use strict'
+
+const z = 7.25;
+console.log(z);
```
wilgert, line 4: You did not round the number. Please carefully read the instructions and improve your solution.

Week1/homework/6-log an array of animals.js
```diff
@@ -0,0 +1,12 @@
+'use strict'
+
+let items = [];
+console.log();
+console.log(items); // empty array
```
wilgert, line 5: very good

Week1/homework/8-Type checker.js
```diff
@@ -0,0 +1,36 @@
+'use strict'
+
+const y = 60;      //1
+const x = 43;
+const color = 'blue';
+const flag = 'italy';
+
+
+if(typeof y === typeof x){                //2
+    console.log(true);
```
wilgert: The idea of this exercise is to log 'SAME TYPE'  or 'Not the same in all these if statements.

Week1/homework/9-Log the remainder.js
```diff
@@ -0,0 +1,13 @@
+'use strict'
```
wilgert: Did you try to run this code? I think it will throw an error.

Week1/homework/4-Log the string.js
```diff
@@ -0,0 +1,19 @@
...
+
+const myString = 'wesam alkfere';
+
+console.log(' the value will be wesam alkfere'); 
+
+console.log(myString);
+
+const myString ='zaandam';
+
+console.log(' has already been declared.. can not a new value');
```
wilgert: Only now I see this comment. So you did try to run the code and got an error. You have to make some changes in order to make it work.

wilgert (changes requested):

Week1/homework/2-Error debugging.js
wesam-k: in single quotes, we can use \ or ` ` ex : console.log(`I'm awesome`);  //or

console.log('I \'m awesome');

Week1/homework/3-Log the number.js
wesam-k: u are right that different between const and let :) u can check it again

Week1/homework/4-Log the string.js
wesam-k: i don't know what is the problem

Week1/homework/5-Round a number and log it.js
wesam-k: ohhh sorry I corrected it but not saved I will do that again

Week1/homework/4-Log the string.js
wesam-k: it's working

wesam-k:
week 1 finish

PR #231, opened 2019-12-14
JS1.Week3.Enwer

EnwerMemet:


Week3/homework/js-exercises/calculateDogAge.js
```diff
@@ -0,0 +1,11 @@
+'use strect'
+// You know how old your dog is in human years, but what about dog years? Calculate it!
+function calculateDogAge(age) {
+   console.log(`Your doggie is  ${age*7}  years old in dog years!`);
```
marcianoviereck, line 4: nice!

Week3/homework/js-exercises/creditCardValidator.js
```diff
@@ -0,0 +1,32 @@
...
+  string.split("").forEach(e => (sum += Number(e)));
+  if (sum < 17) {
+    return false;
+  }
+//   check The final digit must be even
+  if (string.charAt(15) % 2 !== 0) {
+    return false;
+  }
+//   check if it only one type of number
+  const arr = string.split("").filter(e => Number(e) !== sum / 16);
```
marcianoviereck, line 24: Interesting solution:)

Week3/homework/js-exercises/fortuneTeller.js
```diff
@@ -0,0 +1,12 @@
+'use strict'
+function tellFortune(number,partner,location,job){
+return  `In ${time[Math.floor(Math.random()*5)]}, You will be a ${job[Math.floor(Math.random()*5)]} in ${location[Math.floor(Math.random()*5)]}, and married to ${partner[Math.floor(Math.random()*5)]} with ${number[Math.floor(Math.random()*5)]} kids.`
```
marcianoviereck: Ah nice, but does this still works if the arrays are bigger (or smaller than) 5? If there are more (or less) than 5 elements in it?

Week3/homework/js-exercises/fortuneTeller.js
```diff
@@ -0,0 +1,12 @@
...
+}
+let time =[2020,2022,2024,2026,2028];
+let myChildren = [3,4,5,6,7];
+let myWife = ['Fatime','Patigul','Patime','Aypatem','Patemhan'];
+let myLocation = ['Amsterdam','Kashgar','Urumchi','Istanbul','Medina'];
+let myJob = ['Programmer','Teacher','Officeman','Businessman','Poorman'];
```
marcianoviereck: Ah, do you think it's better to use `let` or `const` here :)?

Week3/homework/js-exercises/fortuneTeller.js
```diff
@@ -0,0 +1,12 @@
+'use strict'
+function tellFortune(number,partner,location,job){
+return  `In ${time[Math.floor(Math.random()*5)]}, You will be a ${job[Math.floor(Math.random()*5)]} in ${location[Math.floor(Math.random()*5)]}, and married to ${partner[Math.floor(Math.random()*5)]} with ${number[Math.floor(Math.random()*5)]} kids.`
+}
+let time =[2020,2022,2024,2026,2028];
+let myChildren = [3,4,5,6,7];
+let myWife = ['Fatime','Patigul','Patime','Aypatem','Patemhan'];
```
marcianoviereck: And don't forget to use `s` at the end of the variable name, to show that there are multiple elements in the array

Week3/homework/js-exercises/giveCompliment.js
```diff
@@ -0,0 +1,11 @@
+'use strict'
+function giveCompliment(name){
+   let compliment = [ 'good','awesome','strong','smart','handsome','clever','nice','perfect','kind','something']
+console.log('You are ', compliment[Math.floor(Math.random()*10) ]+','  , name);
```
marcianoviereck: Here the same question, what if the array is bigger (or smaller) than 10? Does this still work?

marcianoviereck (changes requested):
Nice! The homework is looking really good:) I do have some small change requests here and there!

Week3/homework/js-exercises/giveCompliment.js
EnwerMemet: no, it will not work, it will only run till first element.
should i make it work without *10?

Week3/homework/js-exercises/fortuneTeller.js
EnwerMemet: no, it will not work, it will only run till first element.
should i make it work without *5?

marcianoviereck (approved):
Nicely done! I was a bit late with the re-review but now the homework is approved :)

